PR #4, opened 2019-12-04
Add <Nav> , <NavItem> and <NavGroup> components

matthieuh:
Screen Shot 2019-12-04 at 15 47 22


src/components/Nav.tsx
```diff
@@ -0,0 +1,111 @@
...
+  variant,
+  compose
+} from '@modulz/radix-system';
+
+type NavProps = MarginProps & PaddingProps & HeightProps & MinHeightProps;
+
+const styleProps = compose(margin, padding, height, minHeight);
+
+export const Nav = styled('nav')<NavProps>(
+  css({
```
Darkweak: Static style should be in sass file and shouldn't be written as CSS in JS IMO 🤔

src/components/Nav.tsx
```diff
@@ -0,0 +1,111 @@
...
+);
+
+type NavGroupProps = {
+  variant?: 'left' | 'right';
+};
+
+const navGroupStyleProps = {}
+
+export const NavItem = styled('button')<NavItemProps>(
+  css({
```
Darkweak: Same as above IMO

src/components/Nav.story.tsx
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { Box } from './Box';
+import { Nav, NavItem, NavGroup } from './Nav';
+import { Text } from './Text';
+
+function NavStory() {
+  return (
```
Darkweak: Useless bracket function, you can ommit the return then

src/components/Nav.tsx
matthieuh: CSS-in-JS is a wide debate and I don't think that it should be handle in this particular PR since the whole project is handle this way. Feel free to open an issue.

src/components/Nav.story.tsx
matthieuh: In this particular case there is 2 way to write it this way or with an anonymous function and implicit return but for lot of reasons anonymous functions are not really recommended to write a React component.

PR #6, opened 2019-12-04
Fix headings for sizes 4 and 5

paulocfjunior:
@matthieuh, I noticed that it seems we don't have 9 or 10 available for the `fontSize` property. So then I redistributed the sizes to be incremental from 3 to 8, which seems to be our biggest possible value, and equivalent to 72px. Thoughts?

This is the final result after this change:

image

### DIFF
--- a/src/components/Heading.tsx
+++ b/src/components/Heading.tsx
@@ -30,27 +30,27 @@ export const Heading = React.forwardRef<HTMLHeadingElement, HeadingProps>((props
             textIndent: '-.05em',
           },
           1: {
-            fontSize: 5,
+            fontSize: 4,
             letterSpacing: '-.005em',
             textIndent: '-.06em',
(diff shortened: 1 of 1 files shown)


src/components/Heading.tsx
```diff
@@ -30,27 +30,27 @@ export const Heading = React.forwardRef<HTMLHeadingElement, HeadingProps>((props
             textIndent: '-.05em',
           },
           1: {
-            fontSize: 5,
+            fontSize: 4,
```
Darkweak: Better to switch to sass file and generate style in it

matthieuh: CSS-in-JS is a wide debate and I don't think that it should be handle in this particular PR since the whole project is handle this way. Feel free to open an issue.

PR #8, opened 2019-12-05
Add <Input> Component

matthieuh:
Screen Shot 2019-12-05 at 11 05 02


src/components/Input.tsx
```diff
@@ -0,0 +1,93 @@
...
+          '&:disabled': {
+            borderColor: 'gray',
+            color: 'gray',
+            cursor: 'not-allowed',
+          },
+          '&:focus': {
+            border: '2px solid',
+            borderColor: 'black',
+            cursor: 'text',
+          },
```
paulocfjunior: I think we should also have a hover effect here, which (according to the current design on Traefik) will be very similar to this, but with 1px border.

matthieuh: @paulocfjunior Done 👍

matthieuh:
@paulocfjunior is this PR ready to merge or do you have any other feedback?

paulocfjunior:
I think it's good to go :+1:

PR #9, opened 2019-12-06
Add <Chip> component

paulocfjunior:
image

image

image


src/components/Chip.tsx
```diff
@@ -0,0 +1,92 @@
...
+          }
+        },
+      }),
+      variant({
+        size: {
+          0: {
+            fontSize: 0,
+            lineHeight: '20px',
+            letterSpacing: '.032em',
+            paddingX: '4px',
```
matthieuh: PaddingX seems too small regarding the design and the current Traefik dashboard implemention. In Traefik it's 6px. Maybe we should either set it to 8px based on ou styled-system theme having `paddingX: 1` or set it to 10px.

src/components/Chip.tsx
```diff
@@ -0,0 +1,92 @@
...
+            color: 'grays.7',
+            backgroundColor: 'grays.2',
+          },
+          primary: {
+            color: 'primary',
+            backgroundColor: transparentize(0.85, theme.colors.primary),
+          },
+          blue: {
+            color: 'blue',
+            backgroundColor: 'blues.2',
```
matthieuh: I think we should be consistant and use transparentize on the based color for all chip colors. (ex: `backgroundColor: transparentize(0.85, theme.colors.blue)/transparentize(0.85, theme.colors.green)/transparentize(0.85, theme.colors.orange),`

src/components/Chip.tsx
```diff
@@ -0,0 +1,92 @@
...
+          },
+          2: {
+            fontSize: 2,
+            lineHeight: 2,
+            padding: '6px',
+          },
+        },
+      }),
+      variant({
+        truncate: {
```
matthieuh: Maybe we shouldn't keep ellipsis handling there but remote the `maxWidth` from this component. I think that it should be handle in the parent one.
Ex: 
```
<Box maxWidth="150">
  <Chip>Whatever</Chip>
<Box>
```

src/components/Chip.tsx
paulocfjunior: I found that `setLightness` is better for this situation since we may want to place a chip into a colored background and if the color has some transparency, the background may affect the chip color. 
So I replaced `transparentize` for `setLightness` and used it consistently for all variants. Thank you :slightly_smiling_face:

src/components/Chip.tsx
paulocfjunior: Sure, I adjusted the component to use mostly Theme values, it should be more consistent now.

src/theme.ts
```diff
@@ -62,7 +61,7 @@ export const theme = {
     greens: [
       'hsl(175, 60%, 99%)',
       'hsl(175, 60%, 97%)',
-      'hsl(175, 40%, 72%)',
+      'hsl(175, 40%, 88%)',
       'hsl(175, 34%, 64%)',
       'hsl(175, 100%, 33%)',
       'hsl(175, 62%, 40%)',
```
matthieuh: Green colors are not the right one.
I generated a new configuration based on Zeplin:
```
    green: 'hsl(129, 93%, 33%)',
    greens: [
      'hsl(129, 60%, 99%)',
      'hsl(129, 60%, 97%)',
      'hsl(129, 40%, 88%)',
      'hsl(129, 34%, 64%)',
      'hsl(129, 93%, 33%)',
      'hsl(129, 62%, 30%)',
      'hsl(129, 60%, 20%)',
      'hsl(129, 60%, 12%)',
      'hsl(129, 60%, 10%)',
    ],
```

src/components/Chip.tsx
```diff
@@ -0,0 +1,89 @@
...
+          orange: {
+            color: 'orange',
+            backgroundColor: setLightness(0.90, theme.colors.orange),
+          }
+        },
+      }),
+      variant({
+        size: {
+          0: {
+            fontSize: 0,
```
matthieuh: Checking Zeplin the based font-size is 14px (fontSize: 1) instead of  12px (fontSize: 0)

src/theme.ts
matthieuh: I have also seen that we have a lightBlue color.
Here is a configuration for it:
```
lightBlue: 'hsl(192, 69%, 45%)',
    lightBlues: [
      'hsl(192, 82%, 99%)',
      'hsl(192, 80%, 97%)',
      'hsl(192, 78%, 90%)',
      'hsl(192, 76%, 68%)',
      'hsl(192, 69%, 45%)',
      'hsl(192, 82%, 45%)',
      'hsl(192, 67%, 35%)',
      'hsl(192, 67%, 12%)',
      'hsl(192, 67%, 10%)',
    ],
```

src/components/Chip.tsx
```diff
@@ -0,0 +1,89 @@
...
+};
+
+export const Chip = React.forwardRef<HTMLDivElement, ChipProps>((props, ref) => (
+  <BoxPrimitive
+    {...props}
+    ref={ref}
+    css={[
+      css({
+        fontWeight: 600,
+        margin: 1,
```
matthieuh: Components shouldn't outer spacing so margin shoud be removed. The developer will still have the choice to add a margin using the component (ex: ```<Chip margin={1} >```)

src/theme.ts
matthieuh: I made a sketch to make it clearest:
Screen Shot 2019-12-11 at 11 30 13

matthieuh:
Checking storybook Chip in a button is not correctly vertically aligned

src/components/Chip.story.tsx
```diff
@@ -0,0 +1,52 @@
...
+    <Box mb="2" display='flex' style={{flexWrap: 'wrap'}}>
+      <Chip>Default Chip</Chip>
+      <Chip>1</Chip>
+      <Chip>123</Chip>
+      <a href="#">
+        <Chip>Chip in a Link</Chip>
+      </a>
+    </Box>
+    <Box mb="2" maxWidth='150px'>
+      <Chip truncate={true} title="Some very long text that could be collapsed or truncated">
```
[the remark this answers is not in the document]
paulocfjunior: I thought to add the title to the truncated Chip to make the user able to see the full content when hovering the element. Is that valid?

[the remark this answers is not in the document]
matthieuh: Oh yes, in this case, it makes totally sense :)

paulocfjunior:
Results after review changes:

image
image
image

matthieuh:
LGTM

PR #21, opened 2019-12-18
chore: bump to v0.1.1

matthieuh:


paulocfjunior:
Don't we need to commit the `package-lock` too? 
I think you could use `npm version patch` to do the bump, then it updates both the package and package-lock, and also creates the tag.

matthieuh:
> Don't we need to commit the `package-lock` too?
> I think you could use `npm version patch` to do the bump, then it updates both the package and package-lock, and also creates the tag.

@paulocfjunior done :)

PR #26, opened 2019-12-19
fix: Input background should be white

### LINKED ISSUE #23: Input background should be white instead of transparent
Screen Shot 2019-12-19 at 10 48 29
instead of
Screen Shot 2019-12-19 at 10 48 34

paulocfjunior:
Fixes #23.


paulocfjunior:
Already fixed in https://github.com/containous/faency/pull/27

PR #49, opened 2020-02-10
Add Checkbox component

paulocfjunior:
CheckboxComponent


src/components/Checkbox.tsx
```diff
@@ -0,0 +1,37 @@
+import React, { LabelHTMLAttributes } from "react";
```
matthieuh: LabelHTMLAttributes seems not used
```suggestion
import React from "react";
```

paulocfjunior: Thank you!

PR #61, opened 2020-02-21
Add InputTags component

paulocfjunior:
InputTags


packages/faency/src/components/InputTags.tsx
```diff
@@ -0,0 +1,182 @@
...
+  maxTags?: number
+  maxInlineTags?: number
+  renderTag?: RenderTagType
+  renderOption?: RenderOptionType
+  onChange?: (value: string) => void
+  onSubmit?: (value: string) => void
+  onDeleteTag?: (tag: string) => void
+}
+
+export const InputTags = React.forwardRef<HTMLInputElement, InputTagsProps>(
```
matthieuh: Background of this component should be white instead of transparent as the basic input one.

packages/faency/src/components/InputTags.story.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const [options, setOptions] = useState(allOptions)
+  const [submitted, saveSubmit] = useState<string[]>([])
+
+  return (
+    <>
+      <InputTags
+        placeholder="Input with autocomplete"
+        tags={tags}
+        options={options}
+        onChange={v => setOptions(() => allOptions.filter(o => o.includes(v)))}
```
matthieuh: Could be great to fix all lint warnings: "Missing return type on function".

packages/faency/src/components/InputTags.tsx
```diff
@@ -0,0 +1,182 @@
...
+    {option}
+  </SelectItem>
+)
+
+interface InputTagsProps {
+  value?: string
+  placeholder?: string
+  tags?: string[]
+  options?: string[]
+  maxTags?: number
```
matthieuh: `maxTags` seems not used

packages/faency/src/components/InputTags.tsx
```diff
@@ -0,0 +1,182 @@
...
+  maxInlineTags?: number
+  renderTag?: RenderTagType
+  renderOption?: RenderOptionType
+  onChange?: (value: string) => void
+  onSubmit?: (value: string) => void
```
matthieuh: A think that the value of this component should be the array of selected tags instead of the input one:
- `onSubmit` wording should be reserved at the form level and could be renamed `onChange` 
- `onChange` could be renamed `onInputChange` so we can keep handling logic externally

packages/faency/src/components/InputTags.story.tsx
matthieuh: On a UX perspective I think that it could be better to close the popover after user tag selection.

matthieuh: On this example I would expect to see a new tag selecting an option.

packages/faency/src/components/InputTags.tsx
```diff
@@ -0,0 +1,182 @@
...
+    justify-items: flex-end;
+    max-width: 60%;
+  }
+
+  > * {
+    margin: 0 ${theme.space[1]} ${theme.space[1]} 0;
+  }
+`
+
+const SelectContainer = styled(Box)`
```
matthieuh: On a UX/acceesibility perspective the popover should close clicking outside and as the select component having the ability to select the option with keyboard arrows.

packages/faency/src/components/InputTags.story.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const [value, setValue] = useState('')
+  const addTag = (v: string) => v !== '' && setTags(current => (!current.includes(v) ? [...current, v] : current))
+  const handleSubmit = (v: string) => {
+    addTag(v)
+    setValue('')
+  }
+
+  const handleChange = (value: string) => {
+    setValue(value)
+    setOptions(() => allOptions.filter(o => o.includes(value)))
```
matthieuh: Could be great in the example to makes those search case insensitive.

packages/faency/src/components/InputTags.story.tsx
paulocfjunior: Fixed

packages/faency/src/components/InputTags.tsx
paulocfjunior: Sure, fixed

packages/faency/src/components/InputTags.tsx
paulocfjunior: Yes, now that the logic is handled outside, I can remove it. 👍

packages/faency/src/components/InputTags.tsx
paulocfjunior: I agree about the names, I'm going to rename them. 

About returning the array of tags when the input is submitted, I think if we do this, we will lose the use case on the Marketplace where the input is not related to the tags.

I thought we could keep the tags bound to state in the parent component, this way we don't need to send them in a callback, and then I can use the `onChange` listener to determine whether the user has 'submitted' the input by pressing the Enter key.

Does that make sense?

packages/faency/src/components/InputTags.story.tsx
paulocfjunior: This should be the use case for the market place, where there is no correlation between tags and the input.
Because of that, I preferred to show the entered values in a separated set, do you think it's better to do that in some other way?

Issue #36, opened 2020-01-03
Secondary buttons should have a background on hover


matthieuh:
<img width="119" alt="Screen Shot 2020-01-03 at 11 58 45" src="https://user-images.githubusercontent.com/1550192/71720026-72518e00-2e20-11ea-9a7f-3096ccbde08e.png">
instead of
<img width="129" alt="Screen Shot 2020-01-03 at 11 58 53" src="https://user-images.githubusercontent.com/1550192/71720027-7382bb00-2e20-11ea-8bb2-2e45912256f4.png">

PR #70, opened 2020-02-26
feat: add dark mode

matthieuh:
Fixes #70 #36


packages/website/src/docs/select.mdx
```diff
@@ -0,0 +1,124 @@
...
+description: Useful for allowing users to select only one option from a set.
+---
+
+````js live
+<Select>
+  <Option value="option1" label="Option one" />
+  <Option value="option2" label="Option two" />
+  <Option value="option3" label="Option three" />
+</Select>
+```<PropsTable
```
paulocfjunior: From this line until 69, there's no indentation.

PR #77, opened 2020-03-03
fix: enable builds on *@*.*.* tags instead of v*.*.*

matthieuh:
**Reason:** Build is not triggered with tags like this one: `@containous/faency@0.4.0` due to the new monorepo architecture

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,6 +1,6 @@
 {
   "packages": ["packages/*"],
-  "version": "independent",
+  "version": "0.3.3",
   "npmClient": "yarn",
   "useWorkspaces": true
 }


matthieuh:
@ldez Now we are using lerna and yarn workspaces to handle the projects in a monorepo way. Projects are the `website` and the `@containous/faency` lib. So basically now we have to handle the versionning of two internal projects which can have an independent version. To handle that lerna use this kind of version name: `{project}@{version}`. If it's an issue with containous convention I can investigate and maybe setting the projects as dependent we can use one version for both projects. (cf. https://github.com/lerna/lerna/blob/0e9bda769d944e3f0b9218ec1ecfaf63273baf11/commands/publish/index.js#L254)

matthieuh:
@ldez reading lerna doc I found a way to use the same version for all projects inside the repo so keeping `v*.*.*`tags :) I applied changes to the PR.

PR #80, opened 2020-03-04
Add Dropdown and Menu components

paulocfjunior:
Added a separated Menu component once it can be used on other use cases as well.
Also fixed all the unit tests to support the theme context, and added the test run to the CI.

### DIFF
--- a/packages/faency/src/components/DropdownMenu.story.tsx
+++ b/packages/faency/src/components/DropdownMenu.story.tsx
@@ -1,0 +1,44 @@
+import React from 'react'
+import { storiesOf } from '@storybook/react'
+import { Box } from './Box'
+import { DropdownMenu } from './DropdownMenu'
+import {
+  Menu,
+  MenuItem,
+  MenuItemSeparator,
+  MenuItemLabel,
+  MenuItemCheckbox,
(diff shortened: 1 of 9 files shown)


paulocfjunior:
Hey @matthieuh, I tried to reuse the Radix Menu component for the InputTags component, as autocomplete, but I couldn't get the arrow navigation to work at all. Do you have any idea how we could handle that?

matthieuh:
@paulocfjunior I also tried a bit on your PR but I didn't found a way. Btw Could you think of adding docs for all new components?

paulocfjunior:
For some reason, the components `MenuItemRadio`, `MenuItemRadioGroup`, `MenuItemSeparator` and `MenuItemLabel` are not rendering on the Docs website (they work on the Storybook though), and the Menu's `onDismiss` callback is not working at all, neither on their documentation, that describes it as `onClose` but this prop actually doesn't exist on the component.

Trying to find a solution to this, but no luck so far.

paulocfjunior:
@matthieuh I extracted the test changes to another [pull request](https://github.com/containous/faency/pull/90) and the documentation issue with MenuItems is gone, now it doesn't work only on the unit tests.

Issue #74, opened 2020-03-02
Add Switch component


paulocfjunior:
Add the Switch component to the Fæncy library.

Design #73.

PR #88, opened 2020-03-06
feat: add switch component

matthieuh:
@paulocfjunior on this PR I only enable tests on the Switch component since your are fixing the others in #80


packages/faency/src/components/Switch.story.tsx
```diff
@@ -0,0 +1,42 @@
...
+    <>
+      <Box mb="4">
+        <Text as="p">Uncontrolled</Text>
+        <Switch value="row" defaultChecked mr={1} />
+      </Box>
+
+      <Box mb="4">
+        <Text as="p">Uncontrolled and without children</Text>
+        <Switch value="true" defaultChecked />
+      </Box>
```
paulocfjunior: It seems to have no difference between the first and the second case. What is the difference? And what the `value="row"` stand for?

packages/faency/src/components/Switch.tsx
```diff
@@ -0,0 +1,113 @@
...
+      <FakeSwitch />
+    </SwitchWrapper>
+  )
+}
+
+const SwitchWrapper = styled('label')<MarginProps>(
+  css({
+    position: 'relative',
+    display: 'inline-block',
+    width: 6,
```
paulocfjunior: I tried with `width: 5` (and used `space.2` in the Transform for the `:after` pseudo-component) and it seems to have a better look and feel, what do you think?

Width: 6  (Original)
image

Width: 5 (Suggestion)
image

matthieuh: Indeed I'm agree it looks better! :)

packages/faency/src/components/Switch.story.tsx
matthieuh: @paulocfjunior Fixed the unnecessary example. "row" is the value of the input. It's working the exact same way as an checkbox native element: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox

paulocfjunior: Got it, thanks for the link, that was very clarifying

paulocfjunior:
This issue closes #74.

PR #89, opened 2020-03-10
Add Avatar component

### LINKED ISSUE #66: Add User Avatar component
Design #54.

paulocfjunior:
Closes #66.

image

### DIFF
--- a/packages/faency/src/components/Avatar.tsx
+++ b/packages/faency/src/components/Avatar.tsx
@@ -1,37 +1,18 @@
 import React, { useContext, forwardRef } from 'react'
 import { Avatar as AvatarPrimitive, AvatarProps as AvatarPrimitiveProps } from '@modulz/primitives'
 import { ThemeContext } from 'styled-components'
-import useDarkMode from 'use-dark-mode'
 
 type AvatarProps = AvatarPrimitiveProps & {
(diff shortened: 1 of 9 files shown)


paulocfjunior:
@matthieuh the changes you suggested have been applied! Now the Avatar component will show itself on a dark or light variant according to the theme, but if you want to override it, you can pass a `variant` prop with the value `dark` or `light`.

avatar component

packages/faency/src/components/Avatar.tsx
```diff
@@ -0,0 +1,87 @@
+import React, { useContext, forwardRef } from 'react'
+import { Avatar as AvatarPrimitive, AvatarProps as AvatarPrimitiveProps } from '@modulz/primitives'
+import { ThemeContext } from 'styled-components'
+import useDarkMode from 'use-dark-mode'
```
matthieuh: `use-dark-mode` shouldn't be used inside faency. We are trying to keep components stateless. As a workaround if you want always white or always black why not using hexa codes?

paulocfjunior: sure, makes sense!

paulocfjunior:
@matthieuh it's done! 😄

Issue #86, opened 2020-03-05
Add proper tests to InputTags


paulocfjunior:
I had to remove the placeholder test file for this component due to a Jest error on importing the `react-arrow-nav` library.
The InputTags components should have the feature of navigation by arrow keys when a list of options is provided.

Also, there's a possibility to use the Menu component instead of rendering a custom Menu, which would remove the need to use the `react-arrow-nav` lib, but this must be done carefully to keep the visual specifications.

PR #90, opened 2020-03-10
Fix tests and warnings on Docs website

paulocfjunior:
> Removed test for InputTags temporarily due to Jest import error, opened issue #86 to handle this specific error

- Fixed tests by adding theme provider;
- Fixed warning by changing types on PropTypes for Docs website:
- Added `yarn test` script, only runs Fæncy tests for now; and
- Added `yarn test` to the Travis config file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ before_script:
 script:
   - yarn lint
   - yarn build
+  - yarn test
 
 before_deploy:
   - echo "//registry.npmjs.org/:_authToken=\${NPM_API_TOKEN}" >> $HOME/.npmrc 2> /dev/null
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,8 @@ before_script:
 
 script:
   - yarn lint
-  - yarn build
   - yarn test
+  - yarn build
 
 before_deploy:
   - echo "//registry.npmjs.org/:_authToken=\${NPM_API_TOKEN}" >> $HOME/.npmrc 2> /dev/null
--- a/packages/faency/package.json
+++ b/packages/faency/package.json
@@ -36,7 +36,7 @@
     "prepublishOnly": "npm run build",
     "deploy-storybook": "build-storybook",
     "test": "jest",
-    "test:watch": "jest --watch"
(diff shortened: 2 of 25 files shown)


paulocfjunior:
@matthieuh, I also added `yarn test` script to `.travis.yml`, what do you think about it?

packages/faency/package.json
```diff
@@ -36,7 +36,7 @@
     "prepublishOnly": "npm run build",
     "deploy-storybook": "build-storybook",
     "test": "jest",
-    "test:watch": "jest --watch"
+    "test:watch": "jest --watch --no-cache"
```
paulocfjunior: @matthieuh thoughts on this?

packages/website/src/App.js
```diff
@@ -175,7 +175,7 @@ function App({ element, props: appProps }) {
 }
 
 App.propTypes = {
-  element: PropTypes.elementType,
+  element: PropTypes.node,
```
paulocfjunior: Thoughts on this and the other PropTypes changes?

packages/faency/package.json
matthieuh: @paulocfjunior all PR changes looks great but this one is not solving any issue and will just makes test to run slower.

.travis.yml
```diff
@@ -19,6 +19,7 @@ before_script:
 script:
   - yarn lint
   - yarn build
+  - yarn test
```
matthieuh: I would suggest to do yarn test before yarn build. It should go from the faster to the slower and I think that the building step is the slowest task from the 3.

matthieuh: I would suggest to put `yarn test` before `yarn build` since it should go from the fastest tot he slowest and I think that the building task is the slowest one.

packages/website/src/App.js
matthieuh: Look better indead!

packages/faency/package.json
paulocfjunior: Sure, I agree, I'll remove it

.travis.yml
paulocfjunior: Makes sense, I'll change the order

PR #91, opened 2020-03-11
Proposal: Add default value to Description column in Docs

paulocfjunior:
I noticed that we have a possible field called `default` on the `PropsTable`, but it was not used when displaying the items.
I thought we could add it like this, what do you think?

image

If the value has no default, it just doesn't show it, like this:

image

### DIFF
--- a/packages/website/src/components/PropsTable.js
+++ b/packages/website/src/components/PropsTable.js
@@ -33,7 +33,10 @@ export function PropsTable({ data, title = 'Props' }) {
                       <Chip variant="blue">{value.type}</Chip>
                     </Td>
                     <Td>
-                      <Text>{value.description}</Text>
+                      <Text>
+                        {value.description}
+                        {value.default && ` (Default: '${value.default}')`}
+                      </Text>
                     </Td>
                   </Tr>
                 )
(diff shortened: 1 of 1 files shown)


paulocfjunior:
I also had another idea to improve this, if you think the main idea is valid, we could have a kind of a  `RenderDefault` component, that renders each data type accordingly, for example, showing strings with quotes and numbers and boolean in plain text, maybe with another color...

matthieuh:
@paulocfjunior 

> I also had another idea to improve this, if you think the main idea is valid, we could have a kind of a `RenderDefault` component, that renders each data type accordingly, for example, showing strings with quotes and numbers and boolean in plain text, maybe with another color...

That's a great idea! I mean both this PR and the other suggestion. I will validate this PR and we can open an issue to handle the other. We should also start to handle different kind of issues: tasks/proposal/bugs

matthieuh:
@paulocfjunior I let you decide if you want to put default values for all components in this PR or if you want to handle it separately.

paulocfjunior:
I'm going to let this be merged now and create a task for the `DefaultPropRender`.

PR #95, opened 2020-03-13
Add a PropRender component and update Docs to use it

### LINKED ISSUE #94: Create a DefaultProp component to use on Docs
Based on what was discussed on #91, we should have a component to render props values nicely, according to their type.

For now, we can have the support for the following types:
- [x] string
- [x] number
- [x] boolean
- [x] color

paulocfjunior:
Created a component that displays props with different styles and applied to the Documentation's PropTables.

Closes #94.

Supported types:
- string
- number
- boolean
- color

These are all types we use to our default values today.
If no type is provided, it'll detect automatically by testing the value passed as children.

Note: the Color type shows the color as a string and also a square sample of the color.

image


paulocfjunior:
Any suggestions about the colors and styles applied are welcome!

paulocfjunior:
@matthieuh it's done, I moved the PropRender to the website and removed from Faency 👍

PR #100, opened 2020-03-19
Fix incompatibility problem between arrow navigation dependency and Jest

paulocfjunior:
The tests for InputTags were removed due to an incompatibility problem between one of our dependencies and Jest. So I created a component to replace this dependency and removed it from the project. I also added some tests for the InputTags component.


paulocfjunior:
Closes #86.

PR #104, opened 2020-03-31
refactor: upgrade modulz-primitives

matthieuh:
`modulz-primitives` in it's latest version is starting to use `sx` property to styles components using the theme values (the way it should be done using `styled-system`). (cf https://github.com/modulz/radix/pull/219 and https://github.com/modulz/radix/pull/228)

### DIFF
--- a/packages/faency/src/components/Avatar.tsx
+++ b/packages/faency/src/components/Avatar.tsx
@@ -3,7 +3,6 @@ import { Avatar as AvatarPrimitive, AvatarProps as AvatarPrimitiveProps } from '
 import { ThemeContext } from 'styled-components'
 
 export type AvatarProps = AvatarPrimitiveProps & {
-  textColor?: string
   variant?: 'normal' | 'dark' | 'light'
 }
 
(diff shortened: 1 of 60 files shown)


packages/faency/src/components/Avatar.tsx
```diff
@@ -27,7 +26,7 @@ export const Avatar = forwardRef<HTMLSpanElement, AvatarProps>((props, forwarded
               color: '#fff',
               transition: 'all 200ms ease-in-out',
               '*': {
-                color: '#fff !important',
+                color: '#fff',
```
paulocfjunior, line 29: I think we should keep the important here, it was important for this scenario:

image

Then the text can stay white.

matthieuh: For me this is the expected color displayed there. See it as it: "being in dark mode I want to display the color having as constant "white"". And in our dark theme the "white" constant has as value the black color. The issue could be the way we flip colors between light and dark mode but not what's happening there. Moreover we should try to forbid usage of `!important` in a ui lib.

paulocfjunior: Yeah you’re right, I agree.

PR #120, opened 2020-08-04
Fix missing avatar link import and add doc page

matthieuh:


packages/website/src/docs/avata-link.mdx
```diff
@@ -0,0 +1,93 @@
...
+    },
+    alt: {
+      type: 'string',
+      description: 'A text description of the image.',
+    },
+    variant: {
+      type: 'light | dark',
+      description:
+        'The variant to use, it changes the colors and hover effect. The default value is the current theme.',
+    },
```
paulocfjunior, line 29: Should this table include the `href` prop too?

PR #124, opened 2020-09-28
containous -> traefiklabs rebranding

matthieuh:


paulocfjunior (commented):
Found a couple of things that I think are quick wins for us to add to this PR:
1. On the StoryBook in the Nav page and the topbar image is no longer available.
<details>
<summary>:information_source:&nbsp; Screenshot</summary>

Screen Shot 2020-09-29 at 16 38 14
</details>

2. On the Docs, the Nav page shows the correct icon, but it's weird to see it in base64 on the snippet. For this one, I think it would be nice to extract this base64 to a constant or something similar, as we have on the [Avatar page](http://localhost:8000/docs/avatar/).

<details>
<summary>:information_source:&nbsp; Screenshot</summary>

Screen Shot 2020-09-29 at 16 45 41
</details>

3. Something I'm not 100% sure if we should update is the `LICENSE` file line 189 it's written `Copyright 2019 Containous`.

Everything else looks good to me 👌

paulocfjunior (approved):

PR #141, opened 2021-09-29
chore: upgrade @stitches/react to v1.2.1

matthieuh:


paulocfjunior (approved):

moutoum (approved):
Seems good, but why do we need this ?

PR #143, opened 2021-10-01
chore: export CSS type with default config set

paulocfjunior:
This PR aims to make it easier to import the `CSS` interface with all Stitches configuration defined by default.


matthieuh (approved):
LGTM

Issue #153, opened 2021-10-01
Add Elevations variant and component


paulocfjunior:
### Components to receive the Elevation variant
- Card
- Dialog
- DropdownMenu
- Popover

Additionally, we should have an `<Elevation>` component that will add the proper style to the element that goes in the `children` slot.

### Dark theme
![Screen Shot 2021-10-01 at 11 58 58](https://user-images.githubusercontent.com/38889364/135642341-1b572742-4a29-47c5-aed0-ed8f7ef71c27.png)

### Light Theme
![Screen Shot 2021-10-01 at 11 59 05](https://user-images.githubusercontent.com/38889364/135642359-15883a95-20e5-4c08-805e-66130532d929.png)

Issue #148, opened 2021-10-01
Update Input style


paulocfjunior:
![Screen Shot 2021-10-01 at 11 39 57](https://user-images.githubusercontent.com/38889364/135638828-09450aa6-fe4b-4108-af40-7b9d4cd56438.png)

PR #160, opened 2021-10-18
Update input style

seedy:
## Issue

https://github.com/traefik/faency/issues/148

## How to test

- Run storybook
- Check stories:
  - Input
  - TextField
  - IconButton => small impact from [commit](2c32c26bf0514159baf5bd4ef4b478c7a67f80e4)

## Design opinion required on
- Ghost variant rework for Input
- Adornments for Input
  - specific layout when there are icon buttons

Rest follows design preco from the issue

## Code review
Don't hesitate to be picky, I'd like to see learn your design patterns

## Blocked by

Blocked by #153 (must add elevation on focused input) and #162


seedy:
@seedy  TODO: fix native controls positionning.

seedy:
@matthieuh All comment issues resolved.

I fixed some extra issues I found out with extra stories.

Please review changes + have a look at new stories in storybook (Input + TextField)

Thanks in advance

matthieuh (approved):
LGTM

components/Input/Input.tsx
```diff
@@ -178,72 +131,143 @@ const StyledInput = styled('input', {
...
+      },
+      '&::after': {
+        backgroundColor: '$primary',
+        opacity: 0.05,
+      },
+    },
+  },
+
   variants: {
+    disabled: {
```
seedy, line 233: I don't understand why you made `disabled`a variant, rather than a pseudo-class.

Imho, having `:disabled` seems simpler.

components/Input/Input.tsx
```diff
@@ -178,72 +131,143 @@ const StyledInput = styled('input', {
...
+      backgroundColor: '$inputFocusBg',
+    },
+    '&::after': {
+      backgroundColor: '$primary',
+      opacity: 0.15,
+    },
+  },
+
+  '@hover': {
+    '&:hover': {
```
seedy, line 221: You could simply add `:not(:disabled)` to avoid resetting it in `disabled` variant.

components/Input/Input.tsx
```diff
@@ -178,72 +131,143 @@ const StyledInput = styled('input', {
...
+      },
+    },
+  },
+
   variants: {
+    disabled: {
+      true: {
+        opacity: 0.7,
+
+        '&:hover': {
```
seedy, line 238: In case you keep `disabled` variant, shouldn't you wrap `:hover` with `@hover` ?

components/Text/Text.tsx
```diff
@@ -73,19 +73,29 @@ export const Text = styled('span', {
     },
     transform: {
       uppercase: {
-        textTransform: 'uppercase'
+        textTransform: 'uppercase',
```
seedy, line 76: There are many lint changes, what about making sure we all have the same lint rules ?

seedy:
- Input focused invalid has green border, it would be better as red
- Tab interaction is broken because of adornments order. It would be better to make sure start is before the input and end is after in case of tabbable adornments

components/Input/Input.tsx
matthieuh: cf. https://github.com/traefik/faency/pull/160#discussion_r741171123

components/Input/Input.themes.ts
```diff
@@ -1,51 +1,42 @@
...
-    inputHoverBg: 'hsla(68, 81%, 80%, 0.1)',
-    inputHoverBorderGhost: 'hsla(208, 11%, 45%, 0.9)', // inputBorder with opacity
-    inputText: 'hsla(0, 0%, 100%, 0.8)',
-    inputTextFocused: 'hsl(0, 0%, 100%)',
-    inputPlaceholder: 'hsla(0, 0%, 100%, 0.51)',
-    inputDisabledText: 'hsla(0, 0%, 100%, 0.35)',
-    inputDisabledBorder: 'hsla(208, 11%, 45%, 0.7)', // inputBorder with opacity
-    inputInvalidBorder: '$red8',
+  export const getDark: Factory = () => ({
+    inputBg: '#131f2a',
```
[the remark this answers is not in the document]
matthieuh: Fixed

components/Input/Input.themes.ts
```diff
@@ -0,0 +1,42 @@
...
+    inputDisabledText: Property.Color;
+    inputInvalidBorder: Property.Color;
+  };
+
+  type Factory = (primaryColor?: Property.Color) => Colors;
+
+  export const getLight: Factory = () => ({
+    inputBg: '$deepBlue1',
+    inputBorder: '$grayBlue9',
+    inputFocusBg: tinycolor('black').setAlpha(0.15).toHslString(),
```
seedy, line 22: How did you compute this specific value ? I can't find it in the zeplin mockup

components/Input/Input.tsx
```diff
@@ -0,0 +1,439 @@
...
+    position: 'absolute',
+    top: 0,
+    right: 0,
+    bottom: 0,
+    left: 0,
+  },
+
+  '&:focus-visible': {
+    '&::before': {
+      backgroundColor: '$inputFocusBg',
```
seedy, line 212: As far as I understand, you try to apply this value as a shadow, but it is either applied to the whole background - when I set `focus-visible` on the dom node - or not at all.

`StyledInput` now has a transparent background. I think the result is not what you expected.

Issue #145, opened 2021-10-01
Add Switch component


paulocfjunior:
### Dark Theme
![Screen Shot 2021-10-01 at 11 43 06](https://user-images.githubusercontent.com/38889364/135639452-3370e0f3-b3fe-43fe-a6f5-e47ec8ec01e7.png)

### Light Theme
![Screen Shot 2021-10-01 at 11 43 15](https://user-images.githubusercontent.com/38889364/135639427-7fa40143-1c58-45f7-901a-1fea46772139.png)

PR #161, opened 2021-10-18
Add switch component

seedy:
## Issue

#145 

## How to test

- Run storybook
- Check story : Switch

## Design opinion required on
- Disabled state : not part of mockups
- Focus state: not part of mockups. Should we prefer an elevation ?
- Hovered state: not part of mockups. Should we prefer an elevation ?
- Sizes: already existing and updated without mockups

Rest follows design preco from the issue

blocked by https://github.com/traefik/faency/pull/164

### DIFF
--- a/components/Switch/Switch.themes.ts
+++ b/components/Switch/Switch.themes.ts
@@ -1,0 +1,27 @@
+export namespace Theme {
+  type Colors = {
+    switchBackground: string;
+    switchHoverThumb: string;
+    switchThumb: string;
+    switchHoverActiveThumb: string;
+    switchActiveBackground: string;
+  };
+
+  type Factory = (primaryColor: string) => Colors;
+
+  export const getLight: Factory = (primaryColor) => ({
+    switchBackground: 'hsla(0, 0%, 0%, 0.2)',
+    switchHoverThumb: 'hsl(68, 53%, 36%)',
+    switchThumb: 'hsla(0, 0%, 100%, 0.87)',
+    switchHoverActiveThumb: 'hsla(0, 0%, 0%, 0.2)',
+    switchActiveBackground: 'hsl(68, 53%, 36%)',
+  });
+
+  export const getDark: Factory = (primaryColor) => ({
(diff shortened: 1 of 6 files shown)


seedy:
@matthieuh I updated the component theme with global theme changes and simplified the thumb behaviour : when switch is focused or hovered, I apply some elevation variant.

WDYT about the elevation behaviour ?

components/Switch/Switch.tsx
```diff
@@ -0,0 +1,114 @@
...
+          height: THUMB_DIAMETER * LARGE_RATIO,
+          transform: `translateX(${LARGE_RATIO * THUMB_GAP})`,
+
+          '&[data-state="checked"]': {
+            transform: `translateX(${LARGE_RATIO * (ROOT_WIDTH - THUMB_GAP - THUMB_DIAMETER)}px)`,
+          },
+        },
+      },
+    },
+    disabled: {
```
matthieuh, line 92: When a Switch is disabled element should not be focusable to be consistent with buttons and inputs.

components/Switch/Switch.themes.ts
```diff
@@ -0,0 +1,22 @@
+import tinycolor from 'tinycolor2';
+import { Property } from '@stitches/react/types/css';
+export namespace Theme {
+  type Colors = {
```
matthieuh, line 4: type is missing `switchActiveBackground`

matthieuh (changes requested):
I added some request changes but otherwise great work! 🎉

components/Switch/Switch.tsx
seedy: fixed, I replaced the variant by `&:disabled`

components/Switch/Switch.themes.ts
seedy: fixed, I also updated from upstram

matthieuh (approved):
LGTM

paulocfjunior (approved):
LGTM, friendly reminder to export the `Switch` component from the root `index.ts`, otherwise, we won't be able to import it on another app

PR #169, opened 2021-10-27
feat: add navigation components

paulocfjunior:
https://user-images.githubusercontent.com/38889364/139092123-c46e5d50-662e-487b-8579-2a8de823c558.mp4


components/Navigation/Navigation.themes.ts
```diff
@@ -0,0 +1,44 @@
...
+    navBg: 'white',
+    navButtonBg: 'transparent',
+    navButtonHoverBg: 'hsla(0, 0%, 0%, 0.04)',
+    navButtonActiveBg: 'hsla(0, 0%, 0%, 0.04)',
+    navButtonFocusBg: tinycolor(primaryColor).setAlpha(0.05).toString(),
+    navButtonFocusBorder: 'hsl(68, 53%, 48%)',
+    navButtonText: 'hsla(0, 0%, 0%, 0.54)',
+    navButtonHoverText: 'black',
+    navButtonFocusText: 'black',
+    navButtonActiveText: tinycolor(primaryColor).darken(20).toString(),
```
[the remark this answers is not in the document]
paulocfjunior: It should be fixed now would you mind giving another look?

components/Navigation/Navigation.stories.tsx
```diff
@@ -0,0 +1,125 @@
...
+  NavigationItem,
+} from './Navigation';
+import { modifyVariantsForStory } from '../../utils/modifyVariantsForStory';
+import { Badge } from '../Badge';
+import { Text } from '../Text';
+
+export default {
+  title: 'Components/Navigation',
+  component: NavigationDrawer,
+  subComponents: { NavigationContainer, NavigationLink, NavigationButton },
```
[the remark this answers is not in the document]
paulocfjunior: I think it's better to keep the `as` property in the NavigationItem level, I modified the `NavigationLink` to support adornments so if we want all elements to be links, we can use the `NavigationLink` component instead. WDYT?

seedy:
A highly optional point, as I don't know what our expectations would be regarding keyboard a11y interactions for a navigation list:

I think we could follow [keyboard interactions implemented for Tabs by radix-ui](https://www.radix-ui.com/docs/primitives/components/tabs#keyboard-interactions)

It wouldn't bother me if this was done in a following PR, yours is already functional

seedy (approved):
LGTM.

I permit myself a bonus remark.

There are only a few interactions with the keyboard when using the Navigation component. Maybe it could be a good improvement to add some intuitive keyboard interactions when using the navigation list.

I don't know what our expectations would be regarding keyboard interactions, especially with this component.

I think we could find inspiration in [keyboard interactions implemented for Tabs by radix-ui](https://www.radix-ui.com/docs/primitives/components/tabs#keyboard-interactions)

AFAIC, I'd find useful having arrow (down/up) interactions handled, maybe also tab interaction always focusing the active element first, rest seems less important to me.

It wouldn't bother me if this was done in a following PR, yours is already functional.

WDYT ?

components/Navigation/NavigationItem.stories.tsx
```diff
@@ -0,0 +1,88 @@
...
+          <Badge
+            css={{
+              height: '$3',
+              borderRadius: '$pill',
+              fontSize: '$0',
+              fontWeight: 500,
+              color: '$hiContrast',
+            }}
+          >
+            <Text>{args.endAdornment}</Text>
```
matthieuh, line 70: `Text` seems unnecessary and it overrides the font style you put on the `Badge`.

components/Navigation/Navigation.tsx
```diff
@@ -0,0 +1,133 @@
...
+  endAdornment,
+  children,
+  ...restProps
+}: NavigationItemProps): JSX.Element => {
+  const hasStartAdornment = useMemo(() => Boolean(startAdornment), [startAdornment]);
+  const hasEndAdornment = useMemo(() => Boolean(endAdornment), [endAdornment]);
+
+  return (
+    <NavigationItemWrapper as={as} {...restProps}>
+      {hasStartAdornment && <Flex css={{ pr: '$2' }}>{startAdornment}</Flex>}
```
matthieuh, line 118: When a NavigationItem has no adornment padding is inconsistent:
Screen Shot 2021-11-03 at 10 33 48

Actually seems also the case when it has one:
Screen Shot 2021-11-03 at 10 52 11

matthieuh:
I rebased the branch and updated the component to use theme colors.

seedy:
Active state seems broken in `NavigationItem`: background color does not respect mockup

current:

image
image

expected:

image

components/Navigation/Navigation.tsx
```diff
@@ -20,29 +21,66 @@ const baseNavItemCss = css({
...
+  },
+
+  '&:first-child': {
+    mt: 0,
   },
 
   variants: {
     active: {
       true: {
-        backgroundColor: '$navButtonActiveBg',
```
seedy: I think that's the change related to my bug report

components/Navigation/Navigation.tsx
```diff
@@ -71,7 +109,10 @@ export const NavigationDrawer = styled('nav', {
   flexDirection: 'column',
 
   variants: {
-    elevation: elevationVariant,
+    elevation: Object.keys(elevationVariant).map((_, index) => ({
+      ...elevationVariant[index],
+      bc: `$${index < 10 ? '0' : ''}${index}dp`,
```
seedy, line 114: could be refactored into a helper function, I expect it might be used elsewhere soon

seedy (changes requested):

components/Navigation/Navigation.themes.ts
```diff
@@ -5,37 +5,40 @@ export namespace Theme {
...
-    navButtonHoverBg: 'hsla(0, 0%, 0%, 0.04)',
-    navButtonActiveBg: 'hsla(0, 0%, 0%, 0.04)',
-    navButtonFocusBg: tinycolor(primaryColor).saturate(27).lighten(58).setAlpha(0.33).toString(),
-    navButtonFocusBorder: tinycolor(primaryColor).lighten(12).toString(),
-    navButtonText: 'hsla(0, 0%, 0%, 0.54)',
+    navButtonHoverBg: tinycolor('black').setAlpha(0.04).toHslString(),
+    navButtonHoverBg2: 'transparent',
+    navButtonFocusBg: tinycolor('white').setAlpha(0.15).toHslString(),
+    navButtonFocusBg2: '$primary',
+    navButtonFocusBorder: '$neon8',
```
paulocfjunior, line 27: @matthieuh as the theme allows using other primary colors in the provider, the users can, for example, set the primary color to `blue`. Do you think it would make sense to use `neon` here anyways?

components/Navigation/Navigation.tsx
matthieuh: Should now be fixed :)

components/Navigation/Navigation.tsx
matthieuh: I would go on factorizing code when it will be needed instead. Depending on usages factorization would change.

components/Navigation/Navigation.themes.ts
matthieuh: I pushed a commit to make it way easier to get primary color variations. I would like your input on it :)

colors/index.ts
```diff
@@ -1,6 +1,104 @@
...
+  purpleDark,
+  purpleDarkA,
+  slate,
+  slateA,
+  slateDark,
+  slateDarkA,
+  whiteA,
+} from '@radix-ui/colors';
+
+const allColors = {
```
seedy, line 33: `allColors` naming is misleading to me, I would have expected all colors to be part of this object, rather than only our custom colors.

stitches.config.ts
```diff
@@ -316,37 +251,19 @@ export const { styled, css, createTheme, getCssText, globalCss, keyframes, confi
 
 export const utils = config.utils;
 
-export const customColors = (primary: Property.Color) => {
-  const primaryColor = colors[tokenToPropertyName(primary)];
+export const customColors = (primary: PrimaryColor) => {
+  const primaryColor: ColorInfo = {
+    name: primary,
+    value: darkColors[`${primary}9`],
```
seedy, line 257: :warning:  You pass this value to both `darkTheme` and `lightTheme`, shouldn't you switch to `lightColors[...]` for light theme ?

I tested and confirm it breaks some theming related to `primaryColor`, for instance in button:

- Button light theme focused
image

components/Input/Input.themes.ts
```diff
@@ -14,25 +15,25 @@ export namespace Theme {
...
-  type Factory = (primaryColor?: Property.Color) => Colors;
+  type Factory = (primaryColor: ColorInfo) => Colors;
 
-  export const getLight: Factory = () => ({
+  export const getLight: Factory = (primaryColor) => ({
     inputBg: '$deepBlue1',
     inputBorder: '$grayBlue9',
     inputFocusBg: tinycolor('black').setAlpha(0.15).toHslString(),
-    inputFocusBorder: '$neon8',
+    inputFocusBorder: `$${primaryColor.name}8`,
```
seedy, line 24: Sounds a lot like choosing a hue for a color, what about a helper ?
```
type HueValue: '1' | '2' | '3' | '4' | '5' | '6' | '7' | '8' | '9' | '10' | '11' | '12' | 'A1' | A'2' | 'A3' | 'A4' | 'A5' | 'A6' | 'A7' | 'A8' | 'A9' | 'A10' | 'A11' | 'A12'
type Hue: (primaryColor: ColorInfo, value: HueValue) => Property.Color

const hue: Hue = ({name}, value) => `$${name}${value}`
```

components/Card/Card.themes.ts
```diff
@@ -13,14 +14,14 @@ export namespace Theme {
...
-  type Factory = (primaryColor: Property.Color) => Colors;
+  type Factory = (primaryColor: ColorInfo) => Colors;
 
   export const getLight: Factory = (primaryColor) => ({
     cardBackground: 'white',
     cardBorder: '$deepBlue3',
     cardShadow: 'rgba(0,0,0,.1)',
     cardHoverBackground: 'rgba(0,0,0,.05)',
-    cardHoverBorder: tinycolor(primaryColor).setAlpha(0.6).toString(),
+    cardHoverBorder: tinycolor(primaryColor.value).setAlpha(0.6).toString(),
```
seedy, line 24: We could globally replace `tinycolor(...).toString()` by `.toHslString()`

seedy (changes requested):
- 1 comment reporting a bug (:warning: )
- other comments are optional improvements

colors/index.ts
paulocfjunior: Yeah, I agree and think `customColors`, as you mentioned, would be a good name

colors/index.ts
```diff
@@ -1,6 +1,104 @@
...
+  ...deepBlue,
+  ...grayBlue,
+  ...neon,
+  ...orange,
+  ...red,
+};
+
+type LightColors = {
+  [key: string]: string;
+};
```
paulocfjunior, line 44: As both Light and Dark have the same type, and this is not exported, we can have only one type for both, like `Colors` or `ColorMap`

components/Button/Button.themes.ts
```diff
@@ -47,12 +48,12 @@ export namespace Theme {
   export const getDark: Factory = (primaryColor) => ({
     buttonPrimaryBg: '$primary',
     buttonPrimaryText: '$deepBlue2',
-    buttonPrimaryFocusBorder: '$neonA12',
-    buttonPrimaryGhostHoverText: tinycolor(primaryColor).lighten(10).toString(),
+    buttonPrimaryFocusBorder: `$${primaryColor.name}A12`,
+    buttonPrimaryGhostHoverText: tinycolor(primaryColor.value).lighten(10).toString(),
```
paulocfjunior, line 52:
```suggestion
    buttonPrimaryGhostHoverText: tinycolor(primaryColor.value).lighten(10).toHslString(),
```

components/Input/Input.themes.ts
matthieuh: Tbh I'm not finding having this function simplifying it:
```javascript
bc: `$${primaryColor.name}8`
```
vs 

```javascript
import pickHue from 'utils/pickHue';
...
bc: pickHue({ name: primaryColor}, 8),
```

I'm going to find a way to have this instead:

```javascript
bc: pickPrimaryScale(8),
```

matthieuh (approved):
LGTM

utils/getPrimaryColorInfo.ts
```diff
@@ -0,0 +1,28 @@
...
+    pickScale: PickScale;
+  };
+};
+
+const getPrimaryColorInfo = (primaryColor: PrimaryColor, colors: ColorMap): ColorInfo => ({
+  name: primaryColor,
+  value: colors[`${primaryColor}9`],
+  token: `$${primaryColor}9`,
+  helpers: {
+    pickScale: (scale, options = { alpha: false }) =>
```
seedy, line 23: :information_source: This anonymous function is not pure. 
I think it may make sense out of the `getPrimaryColorInfo` scope, so I'd rather make it a pure function that you reuse internally.

```
const pickScale = (primaryColor, scale, options) => ...
```

seedy (approved):
opinionated comment, you can definitely skip it :slightly_smiling_face:

PR #175, opened 2021-11-04
fix: input adornments josep feedback

seedy:
#### Description

On [TextField mockup](https://app.zeplin.io/project/61497d608f9b6e598e281a34/screen/61497eadad40de21822f2aac), Josep recommended to remove any hover/focus interaction for input adornments.

- Replaced IconButton by simple icons with cusor: pointer behaviour on hover
- Removed IconButton integration from input AdornmentWrapper
- Input stories, removed IconButton as adornment

### DIFF
--- a/components/TextField/TextField.tsx
+++ b/components/TextField/TextField.tsx
@@ -4,6 +4,7 @@ import { styled } from '../../stitches.config';
 import { Box } from '../Box';
 import { Input, InputHandle, InputProps, InputVariants } from '../Input';
 import { Label } from '../Label';
+import { Tooltip } from '../Tooltip';
 import {
   ExclamationTriangleIcon,
   CrossCircledIcon,
@@ -142,7 +143,12 @@ export const TextField = React.forwardRef<React.ElementRef<typeof Input>, TextFi
     );
 
     const PasswordVisibilityToggleIcon = React.useMemo(
-      () => isPasswordVisible ? StyledEyeClosedIcon : StyledEyeOpenIcon,
+      () => isPasswordVisible ? StyledEyeOpenIcon : StyledEyeClosedIcon,
+      [isPasswordVisible],
+    );
+
+    const passwordTooltipContent = React.useMemo(
+      () => isPasswordVisible ? "Hide password" : "Show password",
       [isPasswordVisible],
     );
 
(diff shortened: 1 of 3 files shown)


components/TextField/TextField.tsx
```diff
@@ -118,6 +141,11 @@ export const TextField = React.forwardRef<React.ElementRef<typeof Input>, TextFi
       [hasAdornmentGroup]
     );
 
+    const PasswordVisibilityToggleIcon = React.useMemo(
+      () => isPasswordVisible ? StyledEyeClosedIcon : StyledEyeOpenIcon,
```
paulocfjunior, line 145: I feel we should display a closed eye when the password is not visible, and an open eye when the password is visible.

```suggestion
      () => isPasswordVisible ? StyledEyeOpenIcon : StyledEyeClosedIcon,
```

paulocfjunior (approved):
Only one optional feedback, but looks good to me! 👌

components/TextField/TextField.tsx
seedy: @paulocfjunior I expect both cases to be fine.

The matter has never really been solved since timestamps exist : https://ux.stackexchange.com/questions/54162/icon-for-unmasking-passwords-open-or-closed-to-begin-with/54168

I expect the best way would be to add tooltips to make sure the action is understandable.

I don't have a strong opinion on which icon should match which state, though, so I'll apply your recommendation.

paulocfjunior: I agree, didn't know it was such an old discussion. 🤯

matthieuh: I quite agree with Paulo it looks weird to me to have closed eyes when the password is hidden. On the StackExchange link, you are shared I would go on what is said in the message having the most up votes. IMO a button should display a label or icon related to the action it will do.
UX is about user habits so checking what is the common usage is a good way to see what users would expect most. https://dribbble.com/search/password%20input we can see that most designers are putting opened eyes as default.

matthieuh (commented):
I added some water on the opened/closed eyes mill :)

matthieuh (approved):
I added some water on the opened/closed eyes mill but I don't think it should block there PR :)

components/TextField/TextField.tsx
seedy: For the record, @matthieuh is asking the opposite of @paulocfjunior.
To me, the last argument makes sense - follow user habits - plus I'm also used to seeing icon related to the consequence of action rather than current state, when talking about password inputs.
I think the tooltip helps solving the confusion overall. We're used to having the ability to show/hide password, hence the icon won't have a huge impact on the interaction for users.

PR #179, opened 2021-11-05
fix: select style consistency with input one

matthieuh:
# Description

This PR is not about applying the new Select design which will require us to create a custom Select component or wait for radix to provide a Select primitive.
This PR is about having consistency between the Input and the Select component.

Before:
Screen Shot 2021-11-05 at 10 10 50
Screen Shot 2021-11-05 at 10 10 42

After:
Screen Shot 2021-11-05 at 10 10 14
Screen Shot 2021-11-05 at 10 10 05

### DIFF
--- a/components/Select/Select.stories.tsx
+++ b/components/Select/Select.stories.tsx
@@ -29,17 +29,17 @@ export const Basic = Template.bind({});
 
 Basic.args = {};
 
-export const Large = Template.bind({});
+export const Size = Template.bind({});
 
-Large.args = { size: 'large', placeholder: 'placeholder' };
+Size.args = { size: 'large', placeholder: 'placeholder' };
 
-export const Ghost = Template.bind({});
+export const Variant = Template.bind({});
 
-Ghost.args = { variant: 'ghost', value: 'value' };
+Variant.args = { variant: 'ghost', value: 'value' };
 
-export const Invalid = Template.bind({});
+export const State = Template.bind({});
 
-Invalid.args = { state: 'invalid' };
+State.args = { state: 'invalid' };
 
 export const Disabled = Template.bind({});
 
--- a/components/Select/Select.stories.tsx
+++ b/components/Select/Select.stories.tsx
@@ -1,9 +1,10 @@
 import React from 'react';
 import { ComponentStory, ComponentMeta } from '@storybook/react';
 
-import { BaseSelect, SelectProps, SelectVariants } from './Select';
+import { Select, SelectProps, SelectVariants } from './Select';
 import { modifyVariantsForStory } from '../../utils/modifyVariantsForStory';
 
+const BaseSelect = (props: SelectProps): JSX.Element => <Select {...props} />;
 const SelectForStory = modifyVariantsForStory<
   SelectVariants,
   SelectProps & React.InputHTMLAttributes<any>
--- a/components/Select/Select.stories.tsx
+++ b/components/Select/Select.stories.tsx
@@ -17,11 +17,11 @@ export default {
 
 const Template: ComponentStory<typeof SelectForStory> = (args) => (
   <SelectForStory {...args}>
-    <option>Option 1</option>
-    <option>Option 2</option>
-    <option>Option 3</option>
-    <option>Option 4</option>
-    <option>Option 5</option>
+    <option value="option1">Option 1</option>
+    <option value="option2">Option 2</option>
+    <option value="option3">Option 3</option>
+    <option value="option4">Option 4</option>
+    <option value="option5">Option 5</option>
   </SelectForStory>
 );
 
@@ -35,7 +35,7 @@ Size.args = { size: 'large', placeholder: 'placeholder' };
 
 export const Variant = Template.bind({});
 
-Variant.args = { variant: 'ghost', value: 'value' };
+Variant.args = { variant: 'ghost', defaultValue: 'option3' };
 
 export const State = Template.bind({});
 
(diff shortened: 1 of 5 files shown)


components/Select/Select.themes.ts
```diff
@@ -11,35 +12,32 @@ export namespace Theme {
...
+  export const getLight: Factory = (primaryColor) => ({
+    selectBg: '$deepBlue1',
+    selectBorder: '$grayBlue9',
+    selectFocusBg: tinycolor('black').setAlpha(0.15).toHslString(),
+    selectFocusBorder: primaryColor.helpers.pickScale(8),
+    selectHoverBg: '$whiteA9',
+    selectText: tinycolor('black').setAlpha(0.74).toHslString(),
+    selectPlaceholder: '$blackA10',
+    selectDisabledText: tinycolor('black').setAlpha(0.35).toHslString(),
+    selectInvalidBorder: '$red9',
```
seedy, line 29: Simply a suggestion.

It seems we want a strong link between input styles and select styles.

Why copying values, rather than importing them in some way ? I think about either importing them in this theme file, or using them directly in the `Select.tsx` file.

Do you expect this consistency might be unwanted some day ?

We could even remap props if we want to keep both themes distinct or make sure we don't change `$variables` in `Select.tsx`.

While writing a proposal, I notice it's quite annoying and maybe useless, but I hope you get the idea.
```
import InputTheme from '...'

const remapInputTheme = (inputLightOrDark) => {
selectBg: inputLightOrDark.inputBg,
...
}

export const getLight = (primaryColor) => ({
  ...remapInputTheme(InputTheme.getLight(primaryColor),
  // specific select theme colors
})
```

components/Select/Select.stories.tsx
```diff
@@ -1,9 +1,10 @@
...
 import { ComponentStory, ComponentMeta } from '@storybook/react';
 
-import { BaseSelect, SelectProps, SelectVariants } from './Select';
+import { Select, SelectProps, SelectVariants } from './Select';
 import { modifyVariantsForStory } from '../../utils/modifyVariantsForStory';
 
+const BaseSelect = (props: SelectProps): JSX.Element => <Select {...props} />;
 const SelectForStory = modifyVariantsForStory<
   SelectVariants,
   SelectProps & React.InputHTMLAttributes<any>
```
seedy, line 10: I don't know if it's expected, select `ReadOnly` story allows opening dropdown and has no different behaviour (styling + interaction) than `Basic` story:


https://user-images.githubusercontent.com/3367393/140497094-5f2c38c6-8bdb-471c-9024-bb36a8bca52f.mp4


From a storybook user perspective, I'd say the component is broken.


Maybe we could comment this story until it's fixed ?

components/Select/Select.stories.tsx
```diff
@@ -1,9 +1,10 @@
 import React from 'react';
 import { ComponentStory, ComponentMeta } from '@storybook/react';
 
-import { BaseSelect, SelectProps, SelectVariants } from './Select';
+import { Select, SelectProps, SelectVariants } from './Select';
```
seedy, line 4: Stories improvements (a bit unrelated to this PR, feel free to postpone into a new issue):
- stories with `value` prop lack real world interactivity (mainly `Ghost`), could you replace `value` by `defaultValue` ?
- It could be nice to rename `Large` story into `Sizes` and either give a select control for sizes, or display all size variants, I was misled into thinking there's not story for small variant

seedy (approved):
Improvements about stories and a suggestion. I think nothing is blocking, well done :+1:

seedy:
:warning: In dark mode, options text is white on a white background:

https://user-images.githubusercontent.com/3367393/140535307-2f429db3-221a-498f-b325-9cea6db81bb6.mp4

seedy (changes requested):
Dark mode:

Options text is white on a dropdown with white background:

https://user-images.githubusercontent.com/3367393/140535307-2f429db3-221a-498f-b325-9cea6db81bb6.mp4

components/Select/Select.tsx
```diff
@@ -68,6 +72,10 @@ const StyledSelect = styled('select', {
     pointerEvents: 'none',
     color: '$selectDisabledText',
   },
+
+  '&[readonly]': {
```
seedy, line 76: I'd recommend using [pseudo-class selector :read-only](https://developer.mozilla.org/en-US/docs/Web/CSS/:read-only)

components/Select/Select.tsx
```diff
@@ -17,12 +60,26 @@ const StyledSelect = styled('select', {
   pr: '$3',
   lineHeight: '25px',
 
+  '& option': {
+    color: '$loContrast',
```
seedy, line 64: this does not solve dark mode / light mode issue : now in light mode, text has no contrast with the option background:

https://user-images.githubusercontent.com/3367393/140731618-5e80dfbd-5420-45ca-8b4b-75d1e9d0f28f.mp4


The problem is that the select dropdown background does not follow dark/light mode practice : it's always white.

Maybe you should update the select dropdown background when i dark mode, so that it's aligned with the logic.

Here's a working example from mui: https://mui.com/components/selects/#native-select

seedy (changes requested):
Select background does not follow dark/light mode practices, therefore, your change regarding text color for options is now breaking in light mode.

components/Select/Select.tsx
matthieuh: Readonly seems deprecated on the select element so I removed this state. https://html.com/attributes/select-readonly/

seedy (approved):

paulocfjunior (approved):

PR #184, opened 2021-11-09
fix: alpha value on Navigation Item resting state

paulocfjunior:
This PR fixes the alpha value for the NavigationItem in resting state (not active, not focused, not hovered).

Current: 
Screen Shot 2021-11-09 at 12 32 15

Expected: 
Screen Shot 2021-11-09 at 12 32 34


seedy (approved):
tested & reviewed

matthieuh (approved):
LGTM

PR #185, opened 2021-11-10
proposal: set background color based on elevation

paulocfjunior:
Currently, we need to manually set the Background color for the components that use elevation, but we could set it along with the elevation style, which will help to have consistency as we won't need to remember setting the background color anymore.

Please let me know if you agree with the proposition, or if there's a scenario where you think it could become a problem.

### DIFF
--- a/components/Navigation/Navigation.tsx
+++ b/components/Navigation/Navigation.tsx
@@ -126,11 +126,6 @@ export const NavigationContainer = styled('div', {
 export type NavigationContainerVariants = VariantProps<typeof NavigationContainer>;
 export type NavigationContainerProps = NavigationContainerVariants & {};
 
-const elevation = Object.keys(elevationVariant).map((elevationKey) => ({
-  ...elevationVariant[Number(elevationKey)],
-  bc: `$${Number(elevationKey) < 10 ? '0' : ''}${elevationKey}dp`,
-})) as any;
-
 export const NavigationDrawer = styled('nav', {
   display: 'flex',
   boxSizing: 'border-box',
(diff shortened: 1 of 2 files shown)


paulocfjunior:
Demo:

Before:
Screen Shot 2021-11-10 at 04 55 10

Screen Shot 2021-11-10 at 04 55 30

After:
Screen Shot 2021-11-10 at 04 54 17

Screen Shot 2021-11-10 at 04 55 25

seedy (approved):
I lack of context on the implem I guess. I don't expect cases where we don't want both styles applied. It seems like a nice improvement to simplify usage of elevation.

So far, all components using `elevationVariant` lacked of the `bc` (I checked the mockups when they existed):
- `Menu` => we want the `bc`
- `Card` => no mockup
- `Dialog` => no mockup
- `Input` + `Select` when focused => we want the `bc`
- `Popover` => no mockup
- `Switch` when hovered or focused => no mockup

Either we forgot to apply the `bc` or I missed something :confused: 

I'm going to approve, because I think we should have the same behaviour everywhere, anyway :+1:

seedy:
@paulocfjunior :information_source: I think you can remove the extra computation for `bc` in `Navigation` L.115

components/Navigation/Navigation.tsx
```diff
@@ -110,19 +110,14 @@ export const NavigationContainer = styled('div', {
 export type NavigationContainerVariants = VariantProps<typeof NavigationContainer>;
 export type NavigationContainerProps = NavigationContainerVariants & {};
 
-const elevation = Object.keys(elevationVariant).map((elevationKey) => ({
-  ...elevationVariant[Number(elevationKey)],
-  bc: `$${Number(elevationKey) < 10 ? '0' : ''}${elevationKey}dp`,
-})) as any;
-
```
seedy, line 117: @paulocfjunior :information_source: I permitted myself to remove the extra computation for `bc` in `Navigation` L.115

matthieuh (approved):
LGTM

PR #195, opened 2021-11-15
feat: Text component, add inherit variants for fontSize and color

seedy:
### Description

Text component had no `inherit` feature.

- add `variant="inherit"`
- add `size="inherit"`
- updated variant and size stories to display inherit feature

### Preview

image
image

### DIFF
--- a/components/Text/Text.stories.tsx
+++ b/components/Text/Text.stories.tsx
@@ -4,6 +4,7 @@ import { ComponentStory, ComponentMeta } from '@storybook/react';
 import { Text, TextProps, TextVariants } from './Text';
 import { modifyVariantsForStory } from '../../utils/modifyVariantsForStory';
 import { Flex } from '../Flex';
+import { Box } from '../Box';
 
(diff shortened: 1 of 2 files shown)


matthieuh:
I wonder why we would need an `inherit` variant in a design system. Is it not possible to apply `inherit` using CSS?

matthieuh (commented):
I wonder why we would need an inherit variant in a design system. Is it not possible to apply inherit using CSS?

seedy:
@matthieuh 

Because of `defaultVariants` set on the component, it is not possible to create a global component with those styles without using `css` override.

This won't work:
```
const InheritText = styled(Text, {
  fontSize: 'inherit',
  color: 'inherit',
});
```

I thought adding those was a way to reduce code duplication.

My main use case was when combining components.

For instance with Tooltip (for color), Chip - not yet existing though (for fontSize), Buttons with icon + text (for color + fontSize)

paulocfjunior (approved):

matthieuh (approved):
LGTM 👌

PR #197, opened 2021-11-19
Feat: update tooltip styles

seedy:
### Disclaimer

This PR applies styles on tooltips based only on a screenshot. There's no real system design defined for the tooltip component yet.

### Description

- fix css prop ignored by `Tooltip`component
- fix properly passing a ReactNode inside tooltip content: if it's a node, don't wrap it with a `Text`
  - added a story
- add themes file for `Tooltip`, setting content background color and text color

### Found issues

Radix tooltips break around inputs: https://github.com/traefik/faency/issues/211


seedy:
### Disabled children specific case

components/Tooltip/Tooltip.stories.tsx
```diff
@@ -35,3 +67,77 @@ MultiLine.args = {
...
+  content: WarningOption
+};
+
+NodeContent.argTypes = {
+  content: {
+    options: ['Warning', 'Disabled', 'Heading'],
+    mapping: {
+      Warning: WarningOption,
+      Disabled: <Flex align="center" gap={2}><CrossCircledIcon /><Text css={{ color: 'currentColor' }}>Disabled message</Text></Flex>,
+      Heading: <Flex align="center" gap={2}><Text css={{ fontWeight: 700, color: 'CurrentColor' }} >Heading</Text><Text css={{ color: 'currentColor' }}>Content</Text></Flex>,
```
paulocfjunior, line 85: I think it could be great to also move those to variables, it will be more readable IMO.

components/Input/Input.stories.tsx
```diff
@@ -117,40 +118,55 @@ ReadOnly.args = { readOnly: true, defaultValue: 'value' };
...
-      <InputForStory readOnly {...args} />
-    </Label>
-    <Label>
-      Adornments
+    </Box>
+    <Box>
+      <Label htmlFor="default">
+        Default
+      </Label>
+      <InputForStory id="default" {...args} />
```
paulocfjunior, line 131: On the docs part of the storybook, all these stories are shown on the same page, which causes a conflict on the ids, triggering multiple tooltips at once. It doesn't affect the usability of the lib, but I think we should avoid this behavior as our storybook is public.

https://user-images.githubusercontent.com/38889364/143313849-3187fac4-bc90-4c72-a9ec-4fd3f7976b7d.mov


Another example on Tooltips:

https://user-images.githubusercontent.com/38889364/143313406-7fe9b0fc-faf8-4570-bd97-25f6d2f5802f.mov

paulocfjunior:
```suggestion
      <InputForStory id="ghost-default" {...args} />
```

paulocfjunior (changes requested):

components/Input/Input.stories.tsx
seedy: @paulocfjunior This issue happens because `IdProvider` is wrapping each story on its own instead of wrapping the whole docs once. The consequence for tooltips is that `id`s are duplicated between each story.

I create a separate PR to handle this issue

seedy: Fixed for TextField, Label and Input stories: https://github.com/traefik/faency/pull/197/commits/7b40b78294080a5ceb528a7c54ad574860259fbe

:warning: I didn't notice that each story title held an `id` which may conflict with the ids I was setting...
image

@matthieuh @paulocfjunior Any idea how we could ensure we don't fall into this trap again?

matthieuh (changes requested):
Please keep your PRs small doing one thing.

Ex:
- 1 PR for updating `Tooltip` style
- 1 PR per fix
- 1 PR for handling a `disabled` state

Otherwise, It makes it quite difficult to review them.

Please open new PRs splitting the changes.

seedy:
> Please keep your PRs small doing one thing.
> 
> Ex:
> 
> * 1 PR for updating `Tooltip` style
> * 1 PR per fix
> * 1 PR for handling a `disabled` state
> 
> Otherwise, It makes it quite difficult to review them.
> 
> Please open new PRs splitting the changes.

:+1: Done!

seedy:
@matthieuh I split PRs accordingly. You're right, this PR was mixing things up and was difficult to review.
I think the fix specific to tooltip content makes sense in this PR, as it's really small. Please tell me if you want me to move it to another PR.

@paulocfjunior I am sorry :pray: , but I think you'll have to re-review the changes, as I removed half of the commits.

paulocfjunior (approved):
LGTM, just one minor comment 👍

matthieuh (approved):
LGTM

PR #199, opened 2021-11-23
fix: Select, proper variants and props typing

### LINKED ISSUE #198: [Select] size prop type is wrong
### Description

typing:
```
var size: (number & {
    "@motion"?: "small" | "medium" | "large" | undefined;
    "@bp1"?: "small" | "medium" | "large" | undefined;
    "@bp2"?: "small" | "medium" | "large" | undefined;
    "@bp3"?: "small" | "medium" | "large" | undefined;
    ... 4 more ...;
    "@initial"?: "small" | ... 2 more ... | undefined;
} & {
    ...;
}) | undefined
```

`size` prop is passed down to `SelectWrapper` and `StyledCaretSortIcon` which both expect `size` to be a stitches variant, excluding the `number` type.

seedy:
### Description

Rework Select variants and props.

Closes #198

### DIFF
--- a/components/Select/Select.tsx
+++ b/components/Select/Select.tsx
@@ -222,12 +222,12 @@ const SelectWrapper = styled('div', {
   },
 });
 
-export type SelectProps = React.ComponentProps<typeof StyledSelect> &
-  React.ComponentProps<typeof SelectWrapper> & { css?: CSS };
-export type SelectVariants = VariantProps<typeof SelectWrapper>;
+export type SelectVariants = VariantProps<typeof StyledSelect> & VariantProps<typeof SelectWrapper> & VariantProps<typeof StyledCaretSortIcon>;
+
+export type SelectProps = Omit<React.ComponentProps<typeof StyledSelect>, 'size'> & SelectVariants & { css?: CSS };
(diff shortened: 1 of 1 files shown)


paulocfjunior (approved):

matthieuh (commented):
Type fix seems good to me. Regarding the `cursor` change could you explain what you tried to achieve/fix?

components/Select/Select.tsx
```diff
@@ -235,7 +235,7 @@ export const Select = React.forwardRef<React.ElementRef<typeof StyledSelect>, Se
         disabled={props.disabled}
       >
         <StyledSelect ref={forwardedRef} {...props} />
-        <StyledCaretSortIcon size={size} />
+        <StyledCaretSortIcon size={size} cursor={cursor} />
```
seedy, line 238: @matthieuh I binded the `cursor` prop from `Select` down to `StyledCaretSortIcon`, as it has an unused `cursor` variant.

Do you think it's useful? Should I instead remove the variant from the subcomponent?

matthieuh (approved):
LGTM

PR #200, opened 2021-11-23
fix: Select, before and after, inherit borderRadius

seedy:
### Description

Select had slight visible border-radii when hovered, because pseudo elements applying layers of `background-color` didn't have border-radius aligned with parent.

Apply `border-radius: 'inherit'` on both `::before` and `::after` inside `SelectWrapper`

I show before/after with different colors to make it more visible. Note both `::before` and `::after` require the radius because they are layered one of top of the other.

### Before

image

### After

image


paulocfjunior (approved):

matthieuh (approved):
LGTM

PR #201, opened 2021-11-24
chore: export elevationVariants

seedy:
### Description

exporting `elevationVariants` for external custom components to reuse variants when required.

Using `elevationVariants` or `Elevation` component depends on implementation choices Faency shouldn't make.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -18,7 +18,7 @@ export { Card } from './components/Card';
 export { Checkbox } from './components/Checkbox';
 // export { Code } from './components/Code';
 export { Container } from './components/Container';
-export { Elevation } from './components/Elevation';
+export { Elevation, elevationVariants } from './components/Elevation';
 // export { ControlGroup } from './components/ControlGroup';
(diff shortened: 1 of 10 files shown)


index.ts
```diff
@@ -18,7 +18,7 @@ export { Card } from './components/Card';
 export { Checkbox } from './components/Checkbox';
 // export { Code } from './components/Code';
 export { Container } from './components/Container';
-export { Elevation } from './components/Elevation';
+export { Elevation, elevationVariant } from './components/Elevation';
```
paulocfjunior, line 21: WDYT of giving a more meaningful name for this var?

```suggestion
export { Elevation, elevationVariant as elevationStyles } from './components/Elevation';
```

paulocfjunior (approved):
Just one optional comment, but looks good to me 👍

index.ts
seedy: after discussion, @paulocfjunior proposed `elevationVariants`

matthieuh (approved):
LGTM

PR #206, opened 2021-11-25
fix: FaencyProvider, global styles not called properly

seedy:
### Description

This was a console warning. I checked stitches docs: https://stitches.dev/docs/api#globalcss

### DIFF
--- a/components/FaencyProvider.tsx
+++ b/components/FaencyProvider.tsx
@@ -99,7 +99,7 @@ export const FaencyProvider: React.FC<{ primaryColor: PrimaryColor }> = ({
   return (
     <IdProvider>
       <>
-        {globalStyles}
+        {globalStyles()}
         {children}
       </>
     </IdProvider>
--- a/components/FaencyProvider.tsx
+++ b/components/FaencyProvider.tsx
@@ -90,18 +90,18 @@ export const FaencyProvider: React.FC<{ primaryColor: PrimaryColor }> = ({
   children,
   primaryColor,
 }) => {
+
   React.useEffect(() => {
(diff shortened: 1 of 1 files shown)


components/FaencyProvider.tsx
```diff
@@ -43,7 +43,7 @@ export const FaencyProvider: React.FC<{ primaryColor: PrimaryColor }> = ({
   return (
     <IdProvider>
       <>
-        {globalStyles}
+        {globalStyles()}
```
paulocfjunior, line 46: Should we also put it before the return?

seedy: @paulocfjunior I don't understand your question, could you explain?

matthieuh (approved):
LGTM

components/FaencyProvider.tsx
matthieuh: I think @paulo is saying calling `globalStyle` function before the `return` instead of putting it in the jsx (as in the doc https://stitches.dev/docs/api#globalcss).

IMO both are ok. with maybe a preference for putting it before the `return` statement since jsx position will not change anything.

paulo: Yeah, before sounds good 👍

seedy: My bad, I scanned the docs too quickly :joy: 
Thanks for your time folks

matthieuh: Haha thx @paulo joining us as Faency contributors 😂  and sorry for the mention 🙏

paulocfjunior (approved):

Issue #208, opened 2021-11-25
[Storybook] implement uuid generation for stories


seedy:
### Description

Some stories require one or more unique ids to bind inner components together.

For example:
- `Label` stories require to bind `htmlFor` from a `Label` on an `input` `id` prop
- `TextField` stories require a unique `id` prop
- `Input` stories require to bind `htmlFor` from a `Label` on `Input` `id` prop

Radix-UI already implemented an `IdProvider` for its primitives, but it seems not to expose its api.

### Requirements

We need to be able duplicate a story (with `.bind({})`) without duplicating these unique ids.

As we may need multiple unique ids inside of a single story, it would be helpful to be able to configure how many ids we want, and receive them as an array.

### Dependencies

`uuid` and `uuid-browser` are already dependencies of `@storybook/addon-actions`. We could reuse the library if we have no better option.

seedy:
Could be a simple solution: https://github.com/reactwg/react-18/discussions/111

PR #209, opened 2021-11-26
Fix: labels binding in stories

### LINKED ISSUE #210: [Storybook] Input/Label binded stories need unique ids
On the docs part of the storybook, all these stories are shown on the same page, which causes a conflict on the ids, triggering multiple tooltips at once. It doesn't affect the usability of the lib, but I think we should avoid this behavior as our storybook is public.

https://user-images.githubusercontent.com/38889364/143313849-3187fac4-bc90-4c72-a9ec-4fd3f7976b7d.mov

_Originally posted by @paulocfjunior in https://github.com/traefik/faency/pull/197#discussion_r756423779_

seedy:
### Description

Inputs and their Labels are now binded through `htmlFor` + `id` combinations. Hence, they require unique ids inside both "canvas" and "docs" modes.

### Close issue

Closes #210

### Changes

Rework `Input` and `Label` bindings in all related stories:
- Input
- Label
- TextField 

Added a helper `ignoreArgType`  to shorten hiding the `id` arg I had to pass to template stories and their duplicates.

### Misc

:warning: I noticed that each story title held an `id` which may conflict with ids we set inside stories.
image

There's an issue calling for improvements on that subject: #208

### DIFF
--- a/components/Input/Input.stories.tsx
+++ b/components/Input/Input.stories.tsx
@@ -35,28 +35,29 @@ const Template: ComponentStory<typeof InputForStory> = (args) => <InputForStory 
 export const Basic: ComponentStory<typeof InputForStory> = (args) => (
   <Flex direction="column" gap={2}>
     <Box>
-      <Label>Small</Label>
-      <InputForStory size="small" {...args} />
+      <Label htmlFor="small">Small</Label>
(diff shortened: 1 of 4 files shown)


paulocfjunior (approved):

matthieuh (approved):
LGTM

PR #212, opened 2021-11-26
fix: Input, pass DOM node through useImperativehandle

### LINKED ISSUE #211: [Input+Tooltip] `useImperativeHandle` removes access to inner dom element, breaking `Tooltip`
### Description

`Input` component uses `useImperativeHandle` to give parent access to custom features.

`Tooltip` radix primitive requires to call `e.getBoundingClientRect` on wrapped element to properly display around it.
Therefore when it is wrapping an input, its display is broken

seedy:
### Description

Fixed inner `useImperativeHandle` inside `Input`, so that Radix tooltips can properly call `getClientBoundingRect` on them

### Close issues

Closes #211

### DIFF
--- a/components/Input/Input.tsx
+++ b/components/Input/Input.tsx
@@ -367,31 +367,25 @@ export interface HTMLInputProps extends React.InputHTMLAttributes<any> {
 
 export type InputProps = Omit<HTMLInputProps, 'size'> & InputVariants & { css?: CSS };
 
-export type InputHandle = {
+export interface InputHandle extends HTMLInputElement {
   clear: () => void;
-  focus: () => void;
-};
+}
 
 export const Input = React.forwardRef<InputHandle, InputProps>(
   ({ size, startAdornment, endAdornment, css, ...props }, forwardedRef) => {
-    const inputRef = React.useRef<HTMLInputElement>(null);
+    const inputRef = React.useRef<InputHandle>(null);
 
     React.useImperativeHandle(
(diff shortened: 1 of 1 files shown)


components/Input/Input.tsx
```diff
@@ -367,31 +367,25 @@ export interface HTMLInputProps extends React.InputHTMLAttributes<any> {
...
     React.useImperativeHandle(
       forwardedRef,
-      () => ({
-        clear: () => {
-          const { current } = inputRef;
-          if (current) {
-            current.value = '';
-          }
-        },
-        focus: () => {
```
seedy, line 388: :information_source: `focus` is already existing on dom node

components/Input/Input.tsx
```diff
@@ -367,31 +367,25 @@ export interface HTMLInputProps extends React.InputHTMLAttributes<any> {
...
-          const { current } = inputRef;
-          if (current) {
-            current.focus();
-          }
-        },
-      }),
+      () => {
+        const { current } = inputRef;
+        if (current) {
+          current.clear = () => {
```
seedy, line 383: :information_source: I am mutating dom node to add custom features... I'm not really proud to apply mutations here, but the immutable alternative is not working:
```
return {
  ...current, // ignores dom node
  clear: () => }
};
```

paulocfjunior (approved):

matthieuh (approved):
LGTM

PR #213, opened 2021-11-26
Feat: [Tooltip] handle disabled children

seedy:
### Description

PR split from  #197 

## :warning: Accessibility concerns

Because disabled elements are not tabbable, this implementation is not accessible to users without a pointing device.
Therefore, I am not really inclined towards using the feature proposed in this PR.

There is an ongoing dicussion on Radix/Primitives on a weird behaviour regarding Tooltips wrapping inputs: https://github.com/radix-ui/primitives/issues/993
In this issue, maintainers asked information about the use case and gave clear recommendations **not** to offer this behaviour because of its lack of accessibility.
To avoid duplicating references, I'd like to point you towards reading the discussion.

WDYT?

### Design choices

I chose to internally handle the specific case of disabled children, as I expected we might not always remember the requirement from Radix-ui. I guess having a story in storybook for this case is already a good step not to forget.

I thought it would simplify rendering, as we add a node which must be styled as if it didn't exist, not to break the UI. We have `triggerCss` prop to style this extra node properly.

Therefore, the implementation with `trigger="disabled"` might be overkill. **WDYT?**

#### Why passing a prop ?

Checking if the `children` are disabled inside the `Tooltip` component would require lots of opinionated internal computations.

Here's an example implementation, considering a `Tooltip` has maximum 1 child - which should always be true in our current scenario.
```
const [childDisabled, setChildDisabled] = useState(false);

const canCloneChild = (element) => React.isValidElement(element) && typeof element !== 'string'

const composeRef = (element) => (node) => {
  element?.ref(node);
  setChildDisabled(node?.disabled || false);
}

return (
...
{React.Children.map(children, (element) => 
canCloneChild(element) ? React.cloneElement(element, { ref: composeRef(element) }) : element
)}
...
);
```

On each render, we pay the cost of `Children.map`, checking if we can clone the child, cloning the child, rebinding its ref.

I didn't even make sure it was 100% working as there may be edge cases. To me it seems highly overkill.

On the other hand, passing `trigger="disabled"` costs developer understanding.

---

Conclusion: Don't try to check this internally.

#### Styling the extra node by default so that is doesn't break layout

I think it is important to make this behaviour seamless between disabled and not disabled.

I'm looking for the best solution. So far I thought giving the least opinionated solution might help.

Goals:
- don't break flex layout if there's any applied by parent
- inherit parent size not to limit children
- don't apply any extra styling which would be inherited by children


paulocfjunior (approved):
Just one minor comment regarding the code formatting, but LGTM.

seedy:
Closing for now, feature no longer required

PR #215, opened 2021-12-02
feat: overlay add stories

seedy:
### Description

PR on top of #214 

I created stories for Overlay, as they helped me implement aforementioned PR. I think the component brings value in itself, although it is fairly simple. WDYT?

- Adds stories for Overlay component. 
- Improves `overlayStyles` and reworks usage from composing components
  - adds variants
  - composing components use `defaultVariants`

Design is opinionated, I wanted to try a new way to compose styles.
Feel free to give me your opinion or recommendations regarding this choice.

### Preview

https://user-images.githubusercontent.com/3367393/144436158-8eb963b9-e152-4efe-b490-d73ecf5e5a4a.mp4

### DIFF
--- a/components/Overlay/Overlay.tsx
+++ b/components/Overlay/Overlay.tsx
@@ -1,0 +1,18 @@
+import { styled, css } from '../../stitches.config';
+
+export const overlayStyles = css({
+  backgroundColor: 'hsla(0, 0%, 0%, 0.7)',
+  inset: 0,
+  variants: {
+    variant: {
+      fixed: {
+        position: 'fixed',
+      },
+      absolute: {
+        position: 'absolute',
+      }
+    }
+  }
+});
+
+export const Overlay = styled('div', overlayStyles);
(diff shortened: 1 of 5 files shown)


components/Overlay/Overlay.tsx
```diff
@@ -0,0 +1,18 @@
+import { styled, css } from '../../stitches.config';
+
+export const overlayStyles = css({
```
paulocfjunior, line 3: On my tests, I was able to select text behind the Overlay when it was open. I think we should prevent that in some way, maybe using pointer-events on the overlay CSS, or by closing the Overlay when it's clicked. WDYT? 

https://user-images.githubusercontent.com/38889364/144855412-f2d62cba-d3f0-4cdf-9afa-1d41d629fa03.mov

seedy: @paulocfjunior the `Overlay` component is not here to reimplement the `Overlay` features coming from `Dialog`'s overlay.

It's just there to display the styles we apply and make sure those styles, that we reuse are working as expected. Therefore, the feature you request seems unrelated.

Maybe the current story is not clear enough. Please tell me if you have ideas or opinions on how to make clear this story is there simply to display `Overlay` basics without the radix primitive layer

paulocfjunior (approved):

PR #216, opened 2021-12-02
fix: sheet update primitive integration

seedy:
### Description

Sheet component primitive integration was not up to date:
- DialogPrimitive.Close should wrap child and have `asChild`
- IconButton prop `ghost` properly binded

I don't know the point of this component, so far. Should we delete it?

### DIFF
--- a/components/Sheet.tsx
+++ b/components/Sheet.tsx
@@ -123,8 +123,10 @@ export const SheetContent = React.forwardRef<
 >(({ children, ...props }, forwardedRef) => (
   <StyledContent {...props} ref={forwardedRef}>
     {children}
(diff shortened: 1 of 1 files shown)


paulocfjunior:
I'm not sure what it's supposed to do too, but it's not exported from the `index.ts`, neither used from other components.

matthieuh (commented):
I think we should close this PR and open a new one removing all components at the root of the folder `/components' that are not exported in `index.ts` since there are residuals from the project on which we based that one (https://github.com/radix-ui/design-system).

paulocfjunior:
I created an issue to track this need: https://github.com/traefik/faency/issues/224.

PR #217, opened 2021-12-02
fix: Dialog story, proper use of Dialog and DialogTrigger

seedy:
### Description

In Dialog story, example was not properly using Dialog.

I updated the story, based on radix doc: https://www.radix-ui.com/docs/primitives/components/dialog#abstract-the-overlay-and-the-close-button

Consequences: focus is restored to the trigger button when dialog is closed

Test for yourself on the story


paulocfjunior (approved):

matthieuh (approved):
LGTM

PR #231, opened 2021-12-15
feat!: skeleton infer dimensions

### LINKED ISSUE #227: [Skeleton] text variant, infer dimensions from Text leveraging composition
### Overview

For now our Skeleton component doesn't hold the whole complexity of Text variants.

Skeleton sizing is poor.

Maybe it would bring less complexity to compose skeletons for other components by wrapping them inside/around those components, when it makes sense.

### Examples in other libraries

I know of MUI's way to handle skeletons for their typography and would like to recommend something similar:

https://mui.com/components/skeleton/#inferring-dimensions

image

image

seedy:
### Description

Closes #227 

Adds inference for Typographies and Children of Skeleton.

Reworks variants to handle borderRadii of different child UI elements which would be passed as children.

### Preview

https://user-images.githubusercontent.com/3367393/146262318-ac20f9a4-4a1e-4c15-8aa4-11dc7739be32.mp4

#### Some `Skeleton` variants no longer exist

- `avatarX` variants (1 to 6)
- `title`
- `heading`

### DIFF
--- a/components/Skeleton/Skeleton.tsx
+++ b/components/Skeleton/Skeleton.tsx
@@ -37,44 +37,11 @@ export const Skeleton = styled('div', {
 
   variants: {
     variant: {
-      avatar1: {
-        borderRadius: '$round',
-        height: '$3',
-        width: '$3',
+      square: {
+        borderRadius: '$2',
       },
-      avatar2: {
-        borderRadius: '$round',
-        height: '$5',
-        width: '$5',
-      },
-      avatar3: {
-        borderRadius: '$round',
-        height: '$6',
-        width: '$6',
-      },
-      avatar4: {
-        borderRadius: '$round',
-        height: '$7',
-        width: '$7',
-      },
-      avatar5: {
-        borderRadius: '$round',
-        height: '$8',
-        width: '$8',
-      },
-      avatar6: {
-        borderRadius: '$round',
-        height: '$9',
-        width: '$9',
-      },
-      text: {
-        height: '$1',
-      },
-      title: {
-        height: '$5',
-      },
-      heading: {
-        height: '$3',
+      circle: {
+        borderRadius: '$round'
       },
       badge: {
         borderRadius: '$pill',
--- a/components/Skeleton/Skeleton.tsx
+++ b/components/Skeleton/Skeleton.tsx
@@ -10,6 +10,14 @@ export const Skeleton = styled('div', {
   position: 'relative',
   overflow: 'hidden',
   borderRadius: '3px',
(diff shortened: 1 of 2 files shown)


components/Skeleton/Skeleton.tsx
```diff
@@ -30,53 +38,22 @@ export const Skeleton = styled('div', {
...
-        height: '$4',
       },
       button: {
-        borderRadius: '$1',
-        height: '32px',
-        width: '128px',
+        borderRadius: '$3',
+      },
+      text: {
+        '&:empty:before': {
```
seedy, line 54: `empty` seems safe enough for our usage context, but might be clunky in some scenarii:

image


https://developer.mozilla.org/en-US/docs/Web/CSS/:empty

@seedy please add a `FIXME` comment just in case

seedy: Here's a sandbox showing that we should be fine with JSX: https://codesandbox.io/s/kind-ives-4hpeg

components/Skeleton/Skeleton.tsx
```diff
@@ -11,6 +11,14 @@ export const Skeleton = styled('div', {
...
   overflow: 'hidden',
   borderRadius: '3px',
+  height: 'auto',
+  width: 'auto',
+  '&:not(:empty)': {
+    '& > *': {
+      visibility: 'hidden',
+    },
+    maxWidth: 'fit-content',
+  },
```
paulocfjunior, line 21: I'm not sure to get the reason for this part, will we be able to use `Skeleton` around other components? Like
```jsx
<Skeleton>
  <MyComponent />
</Skeleton>
```

paulocfjunior: Nvm, just saw the stories.

components/Skeleton/Skeleton.tsx
```diff
@@ -30,53 +38,22 @@ export const Skeleton = styled('div', {
...
-      avatar4: {
-        borderRadius: '$round',
-        height: '$7',
-        width: '$7',
-      },
-      avatar5: {
-        borderRadius: '$round',
-        height: '$8',
-        width: '$8',
+      square: {
```
paulocfjunior, line 41: ⚠️&nbsp;&nbsp; This is a breaking change if anyone is using `avatar`, `title` or `heading` variants

paulocfjunior (approved):

gndz07 (approved):

seedy:
@seedy Waiting to create a separate release

PR #246, opened 2022-01-18
Fix: label all letters capitalized

### LINKED ISSUE #242: [Label] All letters are capitalize
### Current Behavior

All `Label` letters are capitalize.
Ex: "I'm A Label"

### Expected behavior

Only first letter should be capitalize
Ex: "I'm a label"

### Your environment

| Software         | Version |
| ---------------- | ------- |
| Faency version   |   1.0.0-5    |

seedy:
# Description

Closes #242 

`transform="capitalize"` now capitalizes only first letter.

Adds `transform="capitalizeWords"` in case previous behaviour may be of any use.

Updates stories for `Text` and `Label`

# Warnings

:warning: Added as comment in `Text`: this will only work with block elements (display block/inline-block)

Overriding `Text` or `Label` display might conflict with this variant

[source](https://developer.mozilla.org/en-US/docs/Web/CSS/::first-letter)

### DIFF
--- a/components/Text/Text.tsx
+++ b/components/Text/Text.tsx
@@ -77,6 +77,7 @@ export const Text = styled('span', {
       },
       capitalize: {
         // WARNING: this will only work with block elements (display block/inline-block)
+        // @see https://developer.mozilla.org/en-US/docs/Web/CSS/::first-letter
         '&::first-letter': {
           textTransform: 'uppercase',
         }
--- a/components/Text/Text.tsx
+++ b/components/Text/Text.tsx
@@ -76,8 +76,14 @@ export const Text = styled('span', {
         textTransform: 'uppercase',
       },
       capitalize: {
-        textTransform: 'capitalize',
+        // WARNING: this will only work with block elements (display block/inline-block)
+        '&::first-letter': {
+          textTransform: 'uppercase',
+        }
       },
+      capitalizeWords: {
+        textTransform: 'capitalize',
+      }
     },
     noWrap: {
       true: {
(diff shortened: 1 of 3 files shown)


components/Text/Text.tsx
```diff
@@ -76,8 +76,14 @@ export const Text = styled('span', {
         textTransform: 'uppercase',
       },
       capitalize: {
-        textTransform: 'capitalize',
+        // WARNING: this will only work with block elements (display block/inline-block)
+        '&::first-letter': {
```
seedy, line 80: :question: should I enforce a `display` value here to secure the behaviour ?

components/Text/Text.tsx
```diff
@@ -76,8 +76,15 @@ export const Text = styled('span', {
         textTransform: 'uppercase',
       },
       capitalize: {
-        textTransform: 'capitalize',
+        // WARNING: this will only work with block elements (display block/inline-block)
+        // @see https://developer.mozilla.org/en-US/docs/Web/CSS/::first-letter
```
seedy, line 80: :question: should I enforce a `display` value for this variant to secure the behaviour ?

components/Text/Text.tsx
```diff
@@ -76,8 +76,15 @@ export const Text = styled('span', {
...
+        '&::first-letter': {
```
seedy, line 81: :question: should I enforce a `display` value for this variant to secure the behaviour ?

paulocfjunior: I think you should, yes, `inline-block` as it's a Text element.

paulocfjunior (approved):

components/Text/Text.tsx
seedy: Done. I also updated the default `display` - which was `block` - to `inline-block`

matthieuh (approved):
LGMT

PR #252, opened 2022-01-19
Fix: label capitalize breaking th layout

seedy:
# Description

`Th` extends `Label` styles and variants.

When `transform='capitalize'` default variant is applied to `Th`, style `display: 'inline-block'` takes precedence.
It breaks table layout.

I reset `display: 'table-cell'` from inside `Th` to correct this behaviour.

# Questions

What would you prefer?
1. Removing this enforced display value from `Text`, which could lead to unexpected bugs with `capitalize` when display does not match a block level, but removes the extra logic inside `Th`
2. Doing as I proposed

### DIFF
--- a/components/Table/Table.stories.tsx
+++ b/components/Table/Table.stories.tsx
@@ -66,7 +66,13 @@ Basic.args = {
 Basic.argTypes = {
   transform: {
     control: 'inline-radio',
-    options: ['capitalize', 'capitalizeWords', 'uppercase', '']
+    options: ['capitalize', 'capitalizeWords', 'uppercase', 'none'],
+    mapping: {
+      capitalize: 'capitalize',
+      capitalizeWords: 'capitalizeWords',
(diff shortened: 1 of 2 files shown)


paulocfjunior (approved):

components/Table/Table.stories.tsx
```diff
@@ -60,6 +60,16 @@ export const Basic: ComponentStory<typeof TableForStory> = (args) => (
   </Card>
 );
 
+Basic.args = {
+  transform: 'capitalize',
+};
+Basic.argTypes = {
+  transform: {
+    control: 'inline-radio',
+    options: ['capitalize', 'capitalizeWords', 'uppercase', '']
```
matthieuh, line 69: Radio with empty string looks weird when rendered.

Screen Shot 2022-01-20 at 09 09 24

matthieuh (approved):
Minor feedback other LGTM

PR #257, opened 2022-01-24
feat: radio accordion

seedy:
# Description

Adds `RadioAccordion` component, following mockup

It depends on both `Accordion` and `Radio` styling.

It replaces the default `Chevron` inside `Accordion` with a decoration looking like a radiobutton: it's not an interactive input, it just reproduces the UI of `Radio`.

`RadioAccordion` enforces having a single panel open at a time, plus disables collapsing open panels.

See Radix primitive [Accordion](https://www.radix-ui.com/docs/primitives/components/accordion)

NB: currently `Accordion` exports `AccordionTrigger` as the default UI element, internally rendering an animated chevron. Customizations require exporting `StyledAccordionHeader` and `StyledAccordionTrigger`.

# Preview

https://user-images.githubusercontent.com/3367393/150810532-4ba13408-71c7-40f4-9d5c-372a70a882fd.mp4

# Before you ask

- Why not using a real radiobutton?

The radiobutton UI element would bring extra interactions for the user, but we want none of them.

We simply want an Accordion with more visual feedback than usual.

### DIFF
--- a/components/Radio/Radio.tsx
+++ b/components/Radio/Radio.tsx
@@ -1,13 +1,12 @@
 import React from 'react';
 import * as RadioGroupPrimitive from '@radix-ui/react-radio-group';
 import { styled, CSS, VariantProps } from '../../stitches.config';
-import { modifyVariantsForStory } from '../../utils/modifyVariantsForStory';
 
 export const RadioGroup = styled(RadioGroupPrimitive.Root, {
   display: 'flex',
 });
 
-const StyledIndicator = styled(RadioGroupPrimitive.Indicator, {
+export const INDICATOR_BASE_STYLES = {
   alignItems: 'center',
   display: 'flex',
   height: '100%',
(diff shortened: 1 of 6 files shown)


components/Radio/Radio.tsx
```diff
@@ -49,6 +50,8 @@ const StyledRadio = styled(RadioGroupPrimitive.Item, {
   color: '$hiContrast',
   boxShadow: 'inset 0 0 0 1px $colors$radioBorder',
   overflow: 'hidden',
+}
+const StyledRadio = styled(RadioGroupPrimitive.Item, RADIO_BASE_STYLES, {
```
paulocfjunior, line 54: Just one minor suggestion, the Radio button has a hover effect, do we want to have it on the Accordion too? 

Screen Recording 2022-01-24 at 12 52 54

paulocfjunior (approved):
One optional suggestion, but looks good to me already 👍

components/Radio/Radio.tsx
seedy: @paulocfjunior I think it's a good suggestion to add the same hover effect. To me, it should be applied when hovering over the whole accordion line, to make it clear there's no different behaviour whether you click the "radio decoration" or the rest of the line.

Would it be good for you?

https://user-images.githubusercontent.com/3367393/150834669-a4d75f02-4ee3-4775-9d4d-47c643a5e158.mp4

gndz07 (approved):
LGTM 👍

components/Radio/Radio.tsx
seedy: Crap...

PR #262, opened 2022-01-27
Feat: textfield allow component label

seedy:
# Description

TextField, allow `label` prop to be a Component.

The initial need is to allow nesting more elements than the field's label next to the label, such as tooltips and popovers.

A `ReactNode` could also make sense, but it would lose the bindings to the input internal state, which ensures a good UX.

I created a specific type for the `TextFieldLabelProps`, because they represent the requirement for any label component.

# Preview

https://user-images.githubusercontent.com/3367393/151402668-5179c4b7-7651-4e92-b61a-ac06b69addbb.mp4

### DIFF
--- a/components/TextField/TextField.stories.tsx
+++ b/components/TextField/TextField.stories.tsx
@@ -135,17 +135,19 @@ const StyledInfoCircledIcon = styled(InfoCircledIcon, {
 
 const label = (props) => (
   <Label {...props}>
-    Field Label
-    <Popover>
-      <PopoverTrigger asChild>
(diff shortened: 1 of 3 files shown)


matthieuh (approved):
LGTM

I have just 2 comments:
- Label  is not vertically aligned with the CircledIcon, it looks weirds to me
- Not related to this PR but the `minHeight` on the popover content seems a bad ready that we should remove at some point

wdyt?

gndz07 (approved):
LGTM

PR #265, opened 2022-01-28
Feat!: improve field label behavior

seedy:
# Description

Improve field label behaviour and simplify label color computation.

Until now, `TextField` was applying conflicting rules to its `Label`, for instance:
- `variant: 'red'` + `invalid` when it was invalid
- `variant: 'subtle'` + `disabled` when it was disabled

Expectations where mislead by this and the result did not match them.

# How to test

Check `Label`, `Text`, `TextField`, `Textarea` and `Switch` stories 

# :warning:  Breaking changes

## Text

Reworked text variants related to label behavior inside forms:
- disabled label
- invalid label

Instead of having separate variants, I merged them together inside `variant`

image

Added stories.

## Label

Added stories to follow `Text` changes and display some implementations we made in other places.

# Changes

## TextField

Updated `TextField` label props accordingly

## Textarea

Update `Textarea` label props accordingly

## Switch

I noticed I didn't write stories to show examples of having a Switch with a label.

I wrote 2 stories to follow our mockups: 
image

For example with 2 labels, the behaviour for invalid and disabled states and their compound is a design proposal I made, to try following a similar behaviour to that of `TextField`. I hope we'll have this validated or changed by a mockup one day :pray: 
If you have opinions, I'm up to applying them.

I suppose I could move this part to a separate PR if you prefer. I kept it here simply because I don't add features to the component, but related documentation.

### DIFF
--- a/components/Switch/Switch.stories.tsx
+++ b/components/Switch/Switch.stories.tsx
@@ -34,12 +34,10 @@ export const Disabled = Template.bind({});
 Disabled.args = { disabled: true };
 
 export const Labelled: ComponentStory<typeof SwitchForStory> = ({ id, ...args }) => (
-  <>
-    <Flex align="center">
-      <Label variant="contrast" htmlFor={id}>label</Label>
-      <SwitchForStory id={id} {...args} />
-    </Flex>
-  </>
+  <Flex align="center">
+    <Label variant="contrast" htmlFor={id}>label</Label>
+    <SwitchForStory id={id} {...args} />
+  </Flex>
 );
 Labelled.args = {
   id: 'labelled',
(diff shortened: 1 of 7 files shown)


seedy:
:information_source: Adding no merge as this PR should not be in release 1.0.0-9

components/Switch/Switch.stories.tsx
```diff
@@ -29,3 +32,85 @@ Large.args = { size: '2' };
...
 
 Disabled.args = { disabled: true };
+
+export const Labelled: ComponentStory<typeof SwitchForStory> = ({ id, ...args }) => (
+  <>
+    <Flex align="center">
+      <Label variant="contrast" htmlFor={id}>label</Label>
+      <SwitchForStory id={id} {...args} />
+    </Flex>
+  </>
```
paulocfjunior, line 42: No need for the Fragments here

```suggestion
    <Flex align="center">
      <Label variant="contrast" htmlFor={id}>label</Label>
      <SwitchForStory id={id} {...args} />
    </Flex>
```

paulocfjunior (approved):
LGTM, just one optional improv

matthieuh (approved):
LGTM

PR #268, opened 2022-01-28
fix: radio accordion root, props should not include collapsible

seedy:
# Description

`collapsible` and `type` props should not be allowed, nor required for `AccordionRoot`


gndz07 (approved):

matthieuh (approved):
LGTM

PR #274, opened 2022-02-01
Feat: aria table

seedy:
# Description

- Adds an ARIA-role based table
  - all its structure is made of `div`s and `span`s with proper roles and `display` 
- Adds an unstyled link component
- Adds similar stories as `Table`
- Adds a story showing how to build links as table rows

All recommendations and best practices I read where explicitly telling to avoid using an Aria Table unless other layouts are impossible.

## Use case: each row is an anchor tag

My use case justifying this component over the basic `Table` is to allow having links as table rows.

There are some workarounds as described [here](https://robertcooper.me/post/table-row-links), but in the end, this implementation seems the most accessible outcome for my specific use case.

# Sources

WAI-ARIA
- [wai-aria practices](https://www.w3.org/TR/wai-aria-practices/#table)
- [table example](https://www.w3.org/TR/wai-aria-practices/examples/table/table.html)

MDN
- [table role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/table_role)
- [css display](https://developer.mozilla.org/en-US/docs/Web/CSS/display)

# Preview

https://user-images.githubusercontent.com/3367393/152006653-89221641-6068-4563-b3e4-89371db1a1b4.mp4

# Dependency changes

| Dependency | Version        | State   |
| ---------- | -------------- | ------- |
| [`@radix-ui/react-slot`](https://www.radix-ui.com/docs/primitives/utilities/slot)       | 0.1.2          | Added   |

- [slot](https://www.radix-ui.com/docs/primitives/utilities/slot) allows building components with `asChild` pattern exactly like Radix team does :tada:

### DIFF
--- a/components/AriaTable/AriaTable.stories.tsx
+++ b/components/AriaTable/AriaTable.stories.tsx
@@ -20,10 +20,12 @@ export const Basic: ComponentStory<any> = ({ transform, ...args }) => (
     <TableForStory aria-label="People" aria-describedby="basic-table-caption" {...args}>
       <Caption id="basic-table-caption">People with some information</Caption>
       <Thead>
-        <Th transform={transform}>first name</Th>
-        <Th transform={transform}>last name</Th>
-        <Th transform={transform}>Status</Th>
-        <Th transform={transform}>Role</Th>
+        <Tr>
+          <Th transform={transform}>first name</Th>
+          <Th transform={transform}>last name</Th>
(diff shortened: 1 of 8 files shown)


paulocfjunior (approved):
LGTM 🚀

gndz07 (approved):
LGTM

PR #276, opened 2022-02-02
chore: release v1.0.0-10

seedy:
# Waiting for

- https://github.com/traefik/faency/pull/274
- https://github.com/traefik/faency/pull/237
- https://github.com/traefik/faency/pull/277
- https://github.com/traefik/faency/pull/278 (Breaking change)

# Commits since last release

PR #278, opened 2022-02-03
fix!: avoid init themes on Faency side causing them not being available on server side

matthieuh:
# Description

Themes were initializing on `FaencyProvider` in a `useEffect`  causing them not to be available on the server-side when using with NextJS.

Those changes are now giving full power to people using Faency exporting the themes.

Implementation example using `next-themes`:
Screen Shot 2022-02-03 at 15 40 27


paulocfjunior (approved):

index.ts
```diff
@@ -58,7 +68,8 @@ export {
   globalCss,
   keyframes,
   config,
-  customColors,
+  lightTheme,
+  darkTheme,
```
seedy, line 72: :warning: Breaking changes, please mention them in your PR description

seedy (approved):
LGTM, please make sure breaking changes will be visible for an upcoming release.

Do you think we should add this PR to next release?
=> https://github.com/traefik/faency/pull/276

I added `no-merge` label until a decision is made

matthieuh:
> LGTM, please make sure breaking changes will be visible for an upcoming release.
> 
> Do you think we should add this PR to next release? => #276
> 
> I added `no-merge` label until a decision is made

I think it could be great having it :)

Issue #234, opened 2021-12-17
[Heading] follow `h1-h6` structure for `Heading`


seedy:
# Goals

- make heading usage similar to html heading tags to help usage
- break from `Text` sizing which is unrelated to heading structure
- break from deprecated `as` design pattern #218 

# Context

A Heading is a more complex item than just `font-size`.

They inherently handle other font styles and have a non contiguous size progression.

Headings are ordered from biggest to smallest: `h1` to `h6` for example.

![image](https://user-images.githubusercontent.com/3367393/146525167-89d601ae-2d20-4794-ac79-9bf10728dc97.png)

# Best practices

WAI-ARIA recommends organizing a page using headings and follow a logical hierarchy:
- https://www.w3.org/WAI/WCAG21/Techniques/general/G141
- https://www.w3.org/WAI/WCAG21/Techniques/html/H42

[MDN recommendation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements):
> Avoid skipping heading levels: always start from `<h1>`, followed by `<h2>` and so on.

## Examples in other libraries

- https://mui.com/components/typography/#component
- https://ant.design/components/typography/#components-typography-demo-title

# Proposal

- Remove relation between `Heading` and `Text`
- Reinforce heading structure through variants following html basic structure

PR #279, opened 2022-02-04
feat: rework Heading

seedy:
# Description

Rework `Heading` component

May be impacted by:
- https://github.com/traefik/faency/issues/234


seedy:
:roll_eyes: Tired of handling types :skull:

PR #287, opened 2022-02-10
feat: TextField, ReadonlyCopy story

seedy:
# Description

Adds story example for readonly copy use case.

image

### DIFF
--- a/components/TextField/TextField.stories.tsx
+++ b/components/TextField/TextField.stories.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { ComponentStory, ComponentMeta } from '@storybook/react';
 
 import { Flex } from '../Flex';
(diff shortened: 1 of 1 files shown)


components/TextField/TextField.stories.tsx
```diff
@@ -70,6 +70,40 @@ export const ReadOnly = Basic.bind({});
...
+  }
+});
+
+export const ReadOnlyCopy: ComponentStory<typeof TextFieldForStory> = (args) => {
+  const toCopy = 'Text to copy';
+
+  const onCopy = useCallback(
+    async () => {
+      await navigator.clipboard.writeText(toCopy)
+    },
```
paulocfjunior, line 87: [optional]
I think this could be simplified to just:

```suggestion
    () => navigator.clipboard.writeText(toCopy),
```

paulocfjunior (approved):

components/TextField/TextField.stories.tsx
paulocfjunior: Yeap looks good to me 👍

gndz07 (approved):

PR #295, opened 2022-02-16
feat: Textarea, add endAdornment

seedy:
# Description

Recent mockup required that a `Textarea` allows an `endAdornment`.

image

Adds stories for this example

# Preview

https://user-images.githubusercontent.com/3367393/154340513-a3a62063-2d01-421d-a23d-c271109dca08.mp4

### DIFF
--- a/components/Textarea/Textarea.tsx
+++ b/components/Textarea/Textarea.tsx
@@ -17,7 +17,6 @@ const StyledTextarea = styled('textarea', {
   fontFamily: '$rubik',
   margin: '0',
   outline: 'none',
-  padding: '0',
   WebkitTapHighlightColor: 'rgba(0,0,0,0)',
 
   // Custom
@@ -115,6 +114,12 @@ const StyledTextarea = styled('textarea', {
       },
       horizontal: {
         resize: 'horizontal'
+      }
+    },
+    endAdornment: {
+      true: {
+        paddingBottom: '$5',
+        paddingRight: '$5',
       }
     }
   },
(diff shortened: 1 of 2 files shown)


seedy:
NB: CI is broken while waiting for:
- https://github.com/traefik/faency/pull/294

components/Textarea/Textarea.tsx
```diff
@@ -17,7 +17,6 @@ const StyledTextarea = styled('textarea', {
   fontFamily: '$rubik',
   margin: '0',
   outline: 'none',
-  padding: '0',
```
seedy, line 20: :information_source: removing this reset as we already have `p: '$3'` [here](https://github.com/traefik/faency/pull/295/files#diff-f5d77bd4b99adef81794a7cae30e4c6d288586702603dd0f89daa714b0e2598fR24)

components/Textarea/Textarea.tsx
```diff
@@ -116,6 +115,12 @@ const StyledTextarea = styled('textarea', {
       horizontal: {
         resize: 'horizontal'
       }
+    },
+    endAdornment: {
+      true: {
+        paddingBottom: '$5',
+        paddingRight: '$5',
```
seedy, line 122: :information_source: increase padding when there's an end adornment to make sure it doesn't float on top of textarea content

components/Textarea/Textarea.tsx
```diff
@@ -215,20 +220,35 @@ const TextareaWrapper = styled('div', {
   },
 });
 
-export interface TextareaVariants extends VariantProps<typeof StyledTextarea> { }
+const AdornmentWrapperEnd = styled('div', {
+  position: 'absolute',
+  bottom: '$2',
+  right: '$3',
+  minWidth: '$5',
+  minHeight: '$5',
```
seedy, line 228: :information_source: Expecting max 1 icon with max size of 24px.

seedy:
> @seedy when you have a chance, please update the PR since the linked PR has been merged
> 
> * [fix: kill css incompatible error #294](https://github.com/traefik/faency/pull/294)

@paulocfjunior Done, thanks :+1:

components/Textarea/Textarea.tsx
```diff
@@ -215,20 +220,35 @@ const TextareaWrapper = styled('div', {
...
+  zIndex: 1,
```
paulocfjunior, line 229: > ❓  Question

Will this work in modals?

paulocfjunior (approved):
Just one question, otherwise LGTM 👍

components/Textarea/Textarea.tsx
seedy: @paulocfjunior elements with a `z-index` stack on top of their parent's index.

Therefore, if we have a `Textarea` child of a `Dialog`, it will have the Dialog's z-index + 1.

The only cases where it might be an issue are therefore if we don't have a `zIndex` high enough for elements which should be "on top" of others.

:information_source: Dialogs have the highest z-index, for now, as they're children of a portal.

image

PR #316, opened 2022-03-11
fix: popover arrow color not customizable

### LINKED ISSUE #315: [Popover] arrow color is not customizable
# Description

https://github.com/traefik/faency/blob/master/components/Popover/Popover.tsx#L49

seedy:
## Description

Closes #315 

- Popover arrow customizable
- also export [PopoverAnchor primitive](https://www.radix-ui.com/docs/primitives/components/popover#anchor)
- adds stories

## Good PR checkboxes

- [x] Change has been tested
- [ ] Added/Updated tests
- [x] Added/Updated stories
- [x] Labels are set
- [x] Project is linked

## Good Review checkboxes

<details>

```markdown
- [ ] I've tested the changes
```

</details>

### DIFF
--- a/components/Popover/Popover.stories.tsx
+++ b/components/Popover/Popover.stories.tsx
@@ -74,8 +74,8 @@ export const IconTrigger: ComponentStory<typeof PopoverForStory> = (args) => (
           <HamburgerMenuIcon />
         </Button>
       </PopoverTrigger>
-      <PopoverContent arrowCss={{ color: '$primary' }} css={{ bc: '$primary', p: '$2' }}>
-        <Text css={{ c: 'white' }}>Content</Text>
+      <PopoverContent arrowCss={{ fill: '$primary' }} css={{ bc: '$primary', p: '$2' }}>
+        <Text css={{ c: '$loContrast' }}>Content</Text>
       </PopoverContent>
     </PopoverForStory>
   </Container>
@@ -95,8 +95,8 @@ export const RowAnchor: ComponentStory<typeof PopoverForStory> = (args) => (
           to open the anchor
         </Box>
       </PopoverAnchor>
-      <PopoverContent arrowCss={{ color: '$primary' }} css={{ bc: '$primary', p: '$2' }}>
-        <Text css={{ c: 'white' }}>Content</Text>
+      <PopoverContent arrowCss={{ fill: '$primary' }} css={{ bc: '$primary', p: '$2' }}>
+        <Text css={{ c: '$loContrast' }}>Content</Text>
       </PopoverContent>
     </PopoverForStory>
   </Container>
--- a/components/Popover/Popover.stories.tsx
+++ b/components/Popover/Popover.stories.tsx
@@ -14,6 +14,7 @@ import { Container } from '../Container';
 import { Button } from '../Button';
 import { Text } from '../Text';
 import { Flex } from '../Flex';
+import { HamburgerMenuIcon } from '@radix-ui/react-icons';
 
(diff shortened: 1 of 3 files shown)


components/Popover/Popover.stories.tsx
```diff
@@ -62,3 +65,39 @@ export const RichContent: ComponentStory<typeof PopoverForStory> = (args) => (
...
+export const IconTrigger: ComponentStory<typeof PopoverForStory> = (args) => (
+  <Container>
+    <PopoverForStory {...args}>
+      <PopoverTrigger asChild>
+        <Button size="small">
+          <HamburgerMenuIcon />
+        </Button>
+      </PopoverTrigger>
+      <PopoverContent arrowCss={{ color: '$primary' }} css={{ bc: '$primary', p: '$2' }}>
+        <Text css={{ c: 'white' }}>Content</Text>
```
paulocfjunior, line 78: I think we could maybe use another color/variable here to make it more legible in the dark theme, `$loContrast` here should work 👌 

Screen Shot 2022-03-14 at 12 06 59

components/Popover/Popover.stories.tsx
```diff
@@ -62,3 +65,39 @@ export const RichContent: ComponentStory<typeof PopoverForStory> = (args) => (
...
+      </PopoverContent>
+    </PopoverForStory>
+  </Container>
+);
+
+export const RowAnchor: ComponentStory<typeof PopoverForStory> = (args) => (
+  <Container>
+    <PopoverForStory {...args}>
+      <PopoverAnchor asChild>
+        <Box css={{ bc: '$orange4', p: '$2', display: 'inline-block' }}>
```
paulocfjunior, line 88: I feel we could improve this Story by vertically centralizing the text, and moving it outside the `PopoverForStory` then the Popover will be centralized to the `HamburgerMenuIcon`. 
I also would remove this orange and let the text in `$hiContrast`. WDYT?

This would be the result:

https://user-images.githubusercontent.com/38889364/158204340-63f25483-80bb-4fc8-9916-910c82dfe5e5.mov

components/Popover/Popover.tsx
```diff
@@ -30,28 +30,30 @@ const StyledContent = styled(PopoverPrimitive.Content, panelStyles, {
   },
 });
 
+const StyledArrow = styled(PopoverPrimitive.Arrow, {
+  fill: 'currentColor',
```
paulocfjunior, line 34: I think here we might actually want to pick the same color from `panelStyles.backgroundColor` to avoid this behavior:

Screen Shot 2022-03-14 at 12 24 01

paulocfjunior (changes requested):

components/Popover/Popover.stories.tsx
seedy: @paulocfjunior This example is here to demonstrate the usage of `PopoverAnchor` to anchor the popover to something else than the trigger

components/Popover/Popover.tsx
seedy: I reworked `fill` colors based on `elevationVariants`. It seems to work more consistently

components/Popover/Popover.stories.tsx
paulocfjunior: Humm got it, sounds good.

I'd just keep my recommendation for the text color then, to make it more legible in both color schemes.

paulocfjunior: And it's optional IMO, so I'm approving the PR already.

seedy: done

gndz07 (approved):

PR #319, opened 2022-03-15
feat: aria table colspan

seedy:
## Description

Add `fullColSpan`  boolean prop to `AriaTd`, so that we can span a single column inside a whole table footer. 

Previously, Footer content was not spanning fully, this behavior stayed unnoticed.

After some trials, I ended up with an absolute position hack to ensure the cell always spanned the full footer.

Implementing the `colSpan` behavior is too complex for the need, that's why I went for a hack.

### a11y

a11y-wise, the hidden cell is also hidden from the accessibility tree, as [MDN states](https://developer.mozilla.org/en-US/docs/Web/CSS/visibility#accessibility_concerns). 

For top-notch a11y, it would be nice that implementors precise `aria-colspan`, as shown in the stories.

Computing that value is really tricky on complex tables, hence I'm in favor of letting users setting this attribute themselves. 

### Before

image

### After

image

## How to test?

Check Aria table stories

## Good PR checkboxes

- [x] Change has been tested
- [ ] Added/Updated tests
- [x] Added/Updated stories
- [x] Labels are set
- [x] Project is linked

## Good Review checkboxes

<details>

```markdown
- [ ] I've tested the changes
```

</details>

### DIFF
--- a/components/AriaTable/AriaTable.stories.tsx
+++ b/components/AriaTable/AriaTable.stories.tsx
@@ -1,6 +1,17 @@
 import { ComponentMeta, ComponentStory } from '@storybook/react';
 import React from 'react';
-import { Table, TableProps, TableVariants, Tbody, Td, Th, Thead, Tr, Caption, Tfoot } from './AriaTable';
+import {
+  Table,
+  TableProps,
+  TableVariants,
+  Tbody,
+  Td,
+  Th,
+  Thead,
+  Tr,
+  Caption,
+  Tfoot,
+} from './AriaTable';
 import { Badge } from '../Badge';
 import { Button } from '../Button';
 import { Card } from '../Card';
@@ -63,7 +74,7 @@ export const Basic: ComponentStory<any> = ({ transform, ...args }) => (
       </Tbody>
       <Tfoot>
         <Tr>
-          <Td css={{ textAlign: 'center', columnSpan: 'all' }}>
+          <Td colSpan={4} css={{ textAlign: 'center' }}>
             <Button
               ghost
               variant="secondary"
@@ -90,7 +101,7 @@ Basic.argTypes = {
       capitalizeWords: 'capitalizeWords',
       uppercase: 'uppercase',
       none: '',
-    }
+    },
   },
 };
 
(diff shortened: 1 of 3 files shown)


components/AriaTable/AriaTable.tsx
```diff
@@ -67,8 +68,28 @@ export const Td = forwardRef<
...
-    VariantProps<typeof TableTd> & { css?: CSS }
->((props, ref) => <StyledTd ref={ref} role="cell" {...props} />);
+    VariantProps<typeof TableTd> & { css?: CSS; colSpan?: number }
+>(({ colSpan, css, ...props }, ref) => {
+  const colSpanCss = useMemo(
+    () =>
+      colSpan
+        ? {
+            width: `${Math.max(colSpan || 0, 0) * 100}%`,
+            float: 'left',
```
paulocfjunior, line 78: interesting-megan

paulocfjunior: I don't feel comfortable using `float` but at the same time I didn't find a way to achieve this with `display: table-cell`.

paulocfjunior (approved):

components/AriaTable/AriaTable.tsx
seedy: Yup, this is a tweak...

No documentation on how to achieve the behaviour another way. `column-span: 'all'` doesn't work, it was my best  bet until I found this.

I did a very thorough research before resorting to this solution

components/AriaTable/AriaTable.stories.tsx
```diff
@@ -155,145 +163,207 @@ Alignment.args = {
...
+        <Td>Skywalker</Td>
+        <Td>
+          <Badge variant="red">Disconnected</Badge>
+        </Td>
+        <Td>Star wars</Td>
+      </Tr>
+    </Tbody>
+    <Tfoot>
+      <Tr>
+        <Td colSpan={4} css={{ textAlign: 'center' }}>
```
seedy, line 357: @seedy Please add a story with multiple elements in footer, just to be sure the tweak is not too bad

seedy: Seems not to work as expected... On it

seedy:
After lots of investigation, float layout will only work if there's a single cell within the row.

`float` breaks `display: 'table-cell'` and forces `display: 'block'` instead.

Therefore, this is a hacky solution not to part from table layout and keep in the same time the "span full footer" behavior we need.

I added logic to throw an error in case the `colSpan` feature is not used as expected, to ensure safer usage.

There are only 2 possibles layouts for AriaTable, which are visible in the `Columns` story.

cc @paulocfjunior

components/AriaTable/AriaTable.tsx
paulocfjunior: Yeap, I tried this `column-span` too, no luck. Let's keep it like this until we find a better approach and handle it in another PR.

seedy:
All PRs are locked until resolution of:
- https://github.com/traefik/faency/issues/320

components/AriaTable/AriaTable.tsx
```diff
@@ -63,12 +91,36 @@ export const Th = forwardRef<ElementRef<typeof StyledTh>, ThProps>((props, ref)
...
+  css?: CSS;
+  colSpan?: number;
+}
+export const Td = forwardRef<ElementRef<typeof StyledTd>, TdProps>(
+  ({ colSpan, css, ...props }, ref) => {
+    const colSpanCss = useMemo(
+      () =>
+        colSpan
+          ? {
+              width: `${Math.max(colSpan || 0, 0) * 100}%`,
```
seedy, line 107: this seems to break with `table-layout: 'auto'`, investigating.

image

seedy: `float` hack does not work with `table-layout: 'auto'` because columns don't have the same width.

With absolute positionning, it is mandatory to keep a `visibility: 'hidden'` cell in the footer to ensure the footer does not collapse.

gndz07 (approved):
- [x] I've tested the changes
- [x] I've checked the stories
- [x] I've read the code and understood it

matthieuh (approved):
- [x] I've tested the changes
- [x] I've checked the stories
- [x] I've read the code and understood it
- [x] I don't have any more questions
- [x] I checked PR follows [conventions](https://github.com/traefik/faency#how-to-contribute)

PR #345, opened 2022-04-08
feat!: dialog export back close primitive, export close icon button

seedy:
## Description

- export back the close primitive, as it can be used anywhere inside the dialog as a mean to close it
- export the close icon as a fully styled component

## Preview

No real visual changes, only exports and component revamp

## Breaking changes

- `DialogClose` primitive is back
- added `DialogCloseIconButton` as a separate export component

## How to test?

Review should be enough

## Good PR checkboxes

- [x] Change has been tested
- [ ] Added/Updated tests
- [x] Added/Updated stories
- [x] Labels are set
- [x] Project is linked

## Good Review checkboxes

<details>

```markdown
- [ ] I've tested the changes
```

</details>


gndz07 (approved):

components/Dialog/Dialog.tsx
```diff
@@ -77,13 +91,9 @@ export const DialogContent = React.forwardRef<
...
-        <Cross1Icon />
-      </IconButton>
-    </StyledCloseButton>
+    <DialogCloseIconButton />
   </StyledContent>
 ));
 
 export const DialogTrigger = DialogPrimitive.Trigger;
-export const DialogClose = StyledCloseButton;
+export const DialogClose = DialogPrimitive.Close;
```
paulocfjunior, line 99: I'm wondering if it makes sense to keep `DialogCloseIconButton` named `DialogClose` and call the primitive `DialogClosePrimitive`, wdyt?

As the `DialogCloseIconButton` might be used more often, I think it would be nice to have a simpler/shorter name. I agree that `DialogCloseIconButton` says exactly what the component is at a glance, so feel free to keep it if you prefer it.

paulocfjunior (approved):

components/Dialog/Dialog.tsx
seedy: @paulocfjunior What about having clearer naming rather than misleading names, even if we drop `DialogClose` ?

- I think your proposal of `DIalogClosePrimitive` is clearer than `DialogClose`, thus I'd be in favor of it
- I think `DialogCloseIconButton` is also the clearest naming possible

Issue #218, opened 2021-12-02
`as` design pattern is deprecated


seedy:
### Description

Radix deprecated `as` design pattern and `Polymorphic` in favor of `asChild` design pattern.

Their `Slot` lib component handles internal logic for the `asChild` pattern.

There's [recommendations](https://github.com/radix-ui/primitives/discussions/901) to reuse it, even if they deprecated it to drop support.

I think we should prepare for dropping of `as` design pattern and follow the `asChild` one.

WDYT?

matthieuh:
It seems making sense 👍

paulocfjunior:
Yes, I agree

seedy:
@seedy Handle the update either by listing tasks in issue description or pushing separate issues

Then see if it can be handled quickly or not

seedy:
Related issues:
- https://github.com/traefik/faency/issues/351
- https://github.com/traefik/faency/issues/107

PR #354, opened 2022-04-13
feat: badge `asChild`

### LINKED ISSUE #351: [Badge] "as" prop does not exist on type for Badge
# Description

```
Property 'as' does not exist on type [...]
```

`<Badge as="a"`

seedy:
## Description

- Closes #351 

I chose to implement `asChild` behavior instead of `as` prop, which is really difficult to handle when we implement components on top of `styled`.

Also this choice follows the following RFC:
- https://github.com/traefik/faency/issues/218

usgae of `as` prop is not disabled by this PR, simply, we cannot handle polymorphic typing properly. Users can still remove types from the component and make use of `as` prop if they prefer.

Added test and story to back the use case of a "BadgeLink"

## Preview

image

## Breaking changes

I consider this change non-breaking, because as the issue states, there was an unexpected breaking change brought by https://github.com/traefik/faency/commit/b2851f5c542488e2176185505210db9c75699e5d, which removed `as` prop and  stitches' polymorphic typings. Do you think we should consider this PR bringing a breaking change?

## How to test?

- Check story and test
- Review code
- Share your opinion on whether this PR should be considered breaking change or not

## Good PR checkboxes

- [x] Change has been tested
- [x] Added/Updated tests
- [x] Added/Updated stories
- [x] Labels are set
- [x] Project is linked

## Good Review checkboxes

<details>

```markdown
- [ ] I've tested the changes
```

</details>

### DIFF
--- a/components/Badge/Badge.stories.tsx
+++ b/components/Badge/Badge.stories.tsx
@@ -4,6 +4,7 @@ import { VariantProps } from '../../stitches.config';
 import { Badge, COLORS } from './Badge';
 import { Flex } from '../Flex';
 import { modifyVariantsForStory } from '../../utils/modifyVariantsForStory';
+import { UnstyledLink } from '../Link';
 
 type BadgeVariants = VariantProps<typeof Badge>;
 type BadgeProps = BadgeVariants & {};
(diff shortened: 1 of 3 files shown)


paulocfjunior:
> I consider this change non-breaking, because as the issue states, there was an unexpected breaking change brought by https://github.com/traefik/faency/commit/b2851f5c542488e2176185505210db9c75699e5d, which removed as prop and stitches' polymorphic typings. Do you think we should consider this PR bringing a breaking change?

I think it's not a breaking change. We can see from the stories that all other cases were kept untouched.

paulocfjunior (approved):

matthieuh (approved):

PR #392, opened 2022-12-14
chore: upgrade project to node 18 and react 18

paulocfjunior:
- Upgraded Node to v18
- Upgraded React to v18
- Upgrade Husky and reconfigure hooks
- Set `.nvmrc` for ease of use

### DIFF
--- a/components/AriaTable/AriaTable.tsx
+++ b/components/AriaTable/AriaTable.tsx
@@ -6,8 +6,7 @@ import React, {
   Children,
   isValidElement,
 } from 'react';
-import * as Slot from '@radix-ui/react-slot';
-
+import { Slot } from '@radix-ui/react-slot';
 import { styled, VariantProps, CSS } from '../../stitches.config';
 import {
   Caption as TableCaption,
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -34,44 +34,46 @@
     "prettier": "prettier --write .",
     "build": "rollup -c",
     "release": "npm run build && np",
-    "storybook": "start-storybook -p 6006",
+    "storybook": "NODE_OPTIONS=--openssl-legacy-provider start-storybook -p 6006 -h localhost",
```
gndz07, line 37: Fair warning that this might need a `cross-env` prefix for windows (so it's a warning for me 😀)

gndz07:
I found it weird that it still gives this warning:

image

According to [this](https://github.com/storybookjs/storybook/issues/16797), storybook v6.5 should have been supporting the root API out of the box 🤔 
Do you get the same warning? If it really is behaving as if still running on React v17, I think it won't really solve the rehydration problem on SSR.

gndz07 (approved):
LGTM

PR #397, opened 2023-01-09
feat!: DropdownMenu adjustment

gndz07:
On top of https://github.com/traefik/faency/pull/395

Related to https://github.com/traefik/hub-ui/pull/814

## Breaking Changes

DropdownMenuPortal has to be added as a child of DropdownMenu to trigger portal behavior.


paulocfjunior (changes requested):
LGTM overall, there's however a small blocking issue I found, the Inputs are focusing on the  `::after` pseudo-element instead of the Input element itself, also it seems I'm unable to click on the adornments, because of the same issue.

I wonder if it's related to the `z-index` change applied in `Input.ts`. Maybe we could solve this by disabling `pointer-events` in the pseudo-element, wdyt?

gndz07:
> I wonder if it's related to the `z-index` change applied in `Input.ts`. Maybe we could solve this by disabling `pointer-events` in the pseudo-element, wdyt?

@paulocfjunior  yes! Sorry didn't catch that earlier, apparently it also causes inability to make a text input. Thank you for pointing it out 😁

paulocfjunior (approved):

PR #413, opened 2023-09-04
fix: chevron on accordion trigger should only consider direct parent's data state

gndz07:
## Description

This PR fixes a bug on Chevron direction as AccordionTrigger.

The class selector `[data-state=open] &` will consider any parent element with open data state, this raise a problem when the accordion is a child of another element with `data-state` attribute, e.g. a dialog. When the dialog is open, the chevron direction is always on open data state.

Needed for https://github.com/traefik/hub-ui/pull/1021.

## Preview

Before:

image

After: 

image


mdeliatf (approved):
LGTM!

PR #417, opened 2023-11-22
fix: do not force display as table-row

gndz07:
## Description

For table row with `asChild` set as true, we were forcing the `display: table-row` on it. This could mess up the display for elements that has a custom display value set. This PR removes this behavior to set the `display` value.

## Preview

No visual change, please check the `Links` story under AriaTable to see if there were any undesirable side effects.


mdeliatf (approved):
LGTM!

PR #418, opened 2024-01-09
feat: fullwidth variant for navigation

gndz07:
## Description

Related to https://github.com/traefik/hub-issues/issues/982.

To combine the styles between our current UI and stoplight element seamlessly, we will need to make it possible to make navigation items to take the whole width (visually) on hover and focus.

`fullWidth` props added on the nav drawer and container level.

Tricky part is with navigation tree. I haven't find the most beautiful solution yet, for the moment it works the same, `fullWidth` on drawer and NavigationTreeContainer, and for item that is nesting other elements, user will have to add context about in which level is this element's children nested.

Example:
```
<NavigationDrawer fullWidth>
      <NavigationTreeContainer fullWidth {...args}>
        <NavigationTreeItem {...navigationHandlerProps('one')} label="One" subtitle="/one"> {/** LEVEL 0 */}
          <NavigationTreeItem   -------------------------------------------------------->   {/** LEVEL 1 */}
            {...navigationHandlerProps('one-one')}
            as="a"
            label="One.One"
            subtitle="/one-one"
          />
          <NavigationTreeItem ------------------------------------------------------------> {/** LEVEL 1 */}
            {...navigationHandlerProps('one-two')}
            label="One.Two"
            nestedChildrenLevel={2}
          >
            <NavigationTreeItem ----------------------------------------------------------> {/** LEVEL 2 */}
              {...navigationHandlerProps('one-two-one')}
              startAdornment={<ArchiveIcon />}
              label="One.Two.One"
            />
          </NavigationTreeItem>
        </NavigationTreeItem>
      </NavigationTreeContainer>
    </NavigationDrawer>
```

The `nestedChildrenLevel` should specify in which level the children are. This is only needed if the NavTreeItem has children, that is why I opted to take `nestedChildrenLevel` instead of the element's level itself. But please let me know if you have any input to make it clearer on how to use and makes more sense. Default value is 1.

## Preview

https://github.com/traefik/faency/assets/70909035/577f9056-c4a0-401b-8d0b-b500b0e88429

### DIFF
--- a/components/Navigation/Navigation.stories.tsx
+++ b/components/Navigation/Navigation.stories.tsx
@@ -26,6 +26,9 @@ export default {
     elevation: {
       control: 'number',
     },
+    fullWidth: {
+      control: 'boolean',
+    },
   },
 } as ComponentMeta<typeof NavigationDrawer>;
 
@@ -65,6 +68,7 @@ export const Basic = Template.bind({});
 
 Basic.args = {
   elevation: 1,
+  fullWidth: false,
 };
 
 const AdornmentsTemplate: ComponentStory<typeof NavigationDrawerForStory> = (args) => {
--- a/components/Navigation/Navigation.tsx
+++ b/components/Navigation/Navigation.tsx
@@ -114,6 +114,28 @@ const baseNavItemCss = css({
         },
       },
     },
+    fullWidth: {
+      true: {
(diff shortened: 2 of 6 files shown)


mdeliatf (approved):
LGTM!

PR #420, opened 2024-01-19
fix: inner card background color for light mode

gndz07:
## Description

The current background color for inner card is a white color with adjusted alpha. It makes difference on dark mode, but doesn't really do anything on light mode. This PR fixes the background color for `Card` component with `inner` variant.

Related to https://github.com/traefik/hub-ui/pull/1081.

## Preview

image


mdeliatf (approved):
LGTM!

PR #424, opened 2024-03-08
chore: use yarn on release workflow

gndz07:
## Description

There is no more `package-lock.json` after the last update; therefore, `npm ci` command could not run anymore.


mdeliatf (approved):
LGTM!

PR #429, opened 2024-05-07
fix: add portal to tooltip component

gndz07:
## Description

Tooltip content was not nested inside a portal component, which causes overlay issues at some particular situation.

Needed for https://github.com/traefik/hub-issues/issues/1163#issuecomment-2096377872


mdeliatf (approved):
LGTM!

PR #430, opened 2024-05-13
fix: export separated tooltip components

gndz07:
## Description

Neede to fix the unclickable items inside a tooltip issue described in https://github.com/traefik/hub-issues/issues/1189#issuecomment-2104149811.


mdeliatf (approved):
LGTM!

PR #433, opened 2024-05-28
Fix/dist build

gndz07:
## Description

Stitches did not build components' types correctly after the previous update (https://github.com/traefik/faency/pull/431). This PR added a patch for stitches to avoid mistyping.


mdeliatf (approved):
LGTM!

PR #447, opened 2024-08-23
chore(deps): add missing webpack

mloiseleur:
# Motivation

Yarn v4 says it's required by `babel-loader`.
It seems it was there a year before, see https://github.com/traefik/faency/pull/405

### DIFF
--- a/package.json
+++ b/package.json
@@ -127,8 +127,7 @@
     "tinycolor2": "^1.4.2",
     "typescript": "5.4.5",
     "use-debounce": "9.0.2",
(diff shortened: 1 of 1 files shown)


mloiseleur:
@mdeliatf @gndz07 There is a babel-loader v8.2.5, v8.3.0 and even v9.1.3. I'm unsure if it's appropriate to upgrade, so I have not touched it.

gndz07 (approved):

mdeliatf (approved):
LGTM! Thanks!!

Issue #453, opened 2024-10-11
Missing border on collapsible row


gndz07:
### Current Behavior

Related to https://github.com/traefik/hub-ui/pull/1234#issuecomment-2407584550.

When a collapsible table is nested, like the use case we have on hub UI on access details page, somehow the border between each border disappear, like so:

![image](https://github.com/user-attachments/assets/7805d40e-6942-4bd3-aad7-db2bfc446c71)

From what I've seen so far, all rows here getting a `:last-child` pseudoclass, hence it gets the `border: none` style.

### Expected behavior

The should be borders between rows.

### Reproducible example

This [PR ](https://github.com/traefik/hub-ui/pull/1234) contains a workaround for this bug. Removing the commented lines should re-trigger the bug.

PR #454, opened 2024-10-17
fix: table nested in collapsible row separator

gndz07:
## Description

- Fixes https://github.com/traefik/faency/issues/453

We avoid the last element in every table to have a bottom border by styling its children `<td />`. However, the CSS selector is selecting _all descendants_, and not only direct children, which caused nested table to have no border at all for every last element.

## Preview

Before: 
image

After: 
image


mdeliatf (approved):
LGTM!

PR #457, opened 2024-12-13
feat: side panel

gndz07:
## Description

- Fixes https://github.com/traefik/hub-issues/issues/1544

Create a new side panel component. It is based on dialog components, but designed to be simpler to use.

The component's properties:
- `onOpenChange` (func): required. Function to set the panel to open/close.
- `side`: optional, default to `right`. To set the position of the panel. Available options: right, left, top, bottom.
- `open` (boolean): optional, default to false.
- `noOverlay` (boolean): optional, default to false. To set whether or not the component has an overlay.
- `noCloseButton` (boolean): optional, default to false. It won't have the close ('X') button on the top right if set to right.

## Preview

image

### DIFF
--- a/components/SidePanel/SidePanel.stories.tsx
+++ b/components/SidePanel/SidePanel.stories.tsx
@@ -67,7 +67,7 @@ export const Basic: StoryFn<typeof SidePanel> = (args) => {
         </Box>
       </Box>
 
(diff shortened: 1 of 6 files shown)


mdeliatf (approved):
LGTM! Well done @gndz07 !!

PR #458, opened 2025-01-06
fix: use forwardRef for tooltip component

gndz07:
## Description

When rendering tooltip component, a warning is raised on the console:

`Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?`

This PR fix the Tooltip component, to use `React.forwardRef` instead of a function component.

## Preview

No visual change.

## Dependency changes

| Dependency | Version        | State   |
| ---------- | -------------- | ------- |
| @radix-ui/react-tooltip| 1.0.7 -> 1.1.6          | Updated|

## How to test?

Try the tooltip on the storybook. Make sure that no warning is raised on the console on hovering the component.

### DIFF
--- a/components/Tooltip/Tooltip.tsx
+++ b/components/Tooltip/Tooltip.tsx
@@ -33,10 +33,21 @@ const ArrowBox = styled(Box, {
   color: '$tooltipContentBg',
 });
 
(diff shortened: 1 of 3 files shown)


mdeliatf (approved):
LGTM!

PR #5094, opened 2025-04-02
fix: tooltip content type

gndz07:
## Description

Related to https://github.com/traefik/hub-issues/issues/1827
Follow up to https://github.com/traefik/faency/pull/5093

Fix content type for Tooltip to be able to receive ReactNode correctly.


mdeliatf (approved):
LGTM!

PR #5096, opened 2025-07-08
feat!: improve build with vite

gndz07:
## Description

Replace build script with Vite, remove unused rollup dependencies.

Vite is also using rollup under the hood. Build time using vite is considerably faster compared to rollup.

No breaking change. The output should have a similar structure.

However, the release will be a new major release, as the change is quite pertinent and it might mess up with consumer project's typings. So, upgrade with cautions 😄 

## Preview

No visual change.

## How to test?

Please try to build the package locally, and try to link the built package to another project that's using faency. There should be no changes necessary in the other project related to faency. The `dist` output should look like this:
image

### DIFF
--- a/tsconfig-build.json
+++ b/tsconfig-build.json
@@ -25,6 +25,14 @@
       "react": ["./node_modules/@types/react"]
     }
   },
-  "exclude": ["node_modules", "pages", "**/*.test.tsx"],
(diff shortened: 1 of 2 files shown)


mdeliatf (approved):
LGTM!

Issue #422, opened 2024-03-06
[STUDY] migrate from stitches


gndz07:
### Overview

Stitches is [no longer actively maintained](https://github.com/stitchesjs/stitches/discussions/1149#discussioncomment-6223090) since June 2023. It is not a problem for the moment, but it might pose problems if we want to keep up with the ecosystem in the future (for example, if we wanted to use Next.js app folder, stitches wouldn't play so well with it).

We might want to consider to start migrating from stitches to another alternative.

Related discussion: https://github.com/stitchesjs/stitches/discussions/1149

### Additional context

Possible alternatives:
- [Panda](https://panda-css.com/docs/overview/getting-started)
- [cva](https://cva.style/docs)
- [Linaria](https://github.com/callstack/linaria)

gndz07:
## Notes from discussion 06/03

- vanilla-extract is out of context of discussion for now
- Panda seems the easiest to migrate to - similar syntax, some shared functionalities
- cva only has variants as the plus point, on other points Pandas still preferable
- Linaria seems like it doesn't have the necessary functionalities that we'd need

## Next steps

- dig deeper into Panda, identify any possible pain points to migrate
- rediscuss and define tasks to migrate

gndz07:
[StyleX](https://stylexjs.com/) could be another choice too. We are doing a PoC on it.

gndz07:
[Notion document](https://www.notion.so/containous/Alternatives-to-Stiches-for-faency-25707ac78a94800fad71e4166d18fd6e) which contains initial analysis on different alternatives. Until this moment, I went deep mostly on 3 options: Panda, vanilla-extract, and Linaria. So far vanilla-extract is my top choice. Further details in the document.

PR #5100, opened 2025-10-14
chore: kickstart vanilla-extract migration

gndz07:
## Description

Fixes https://github.com/traefik/faency/issues/422

This PoC is part of the study, and we can move forward from this point. 

We decided to choose vanilla-extract for several practical reasons:
- not very different syntax with our current implementation with stitches
- allows us to migrate progressively
- no compiling needed on the user's side

Key points:

- We will migrate progressively, one component after another
- Stitches and vanilla-extract component can live side by side, so we can remove stitches support only after we finish the migration for all of the components
- We aim for no API changes from the user perspective

Please refer to the migration plan and developer guide to get more insight on how we will proceed with this migration.

Two components are migrated already to be used as examples: Box and Badge component.

## Preview

Stitches and vanilla-extract components should be visually identical with the exact same props for each components:

image

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -1,5 +1,5 @@
-import { vanillaExtractPlugin } from '@vanilla-extract/vite-plugin';
 import type { StorybookConfig } from '@storybook/react-vite';
+import { vanillaExtractPlugin } from '@vanilla-extract/vite-plugin';
 
 const config: StorybookConfig = {
   stories: [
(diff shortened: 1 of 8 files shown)


mdeliatf (approved):
LGTM! Amazing work, Gina!!

I just have one non-blocking question: in `CLAUDE.md` we mention that Stitches is the core styling system; shouldn’t we say instead that it was, and that we’re migrating to Vanilla Extract with its specs?

PR #5111, opened 2025-12-08
fix: storybook initial theme for vanilla provider

gndz07:
## Description

On the initial load, vanilla extract provider always fallback to system theme if the forced theme is light.

This PR attempts a fix by setting the right condition check on the initial theme load, loading the theme class immediately to avoid flash, and to use `useLayoutEffect` to avoid repaint.

## How to test?

Check on any comparison stories, on the initial load the theme color for both components (stiches & vanilla) should be the same.


mdeliatf (approved):
LGTM

PR #5133, opened 2026-04-15
feat: add CodeBlock component and fix QA findings

mdeliatf:
## Summary

Introduces a new `CodeBlock` component with syntax highlighting and copy-to-clipboard support, alongside several QA-driven fixes across Button, List, AriaTable, and the build config.

### New: CodeBlock component
- Both Stitches (`CodeBlock`) and vanilla-extract (`CodeBlockVanilla`) implementations
- Syntax highlighting via `prism-react-renderer` with auto light/dark theme switching
- Optional copy button with configurable `copyText`/`copiedText` labels and 1.5 s revert
- `noBorder`, `wrapText`, `css`, and `style` prop support; ref-forwarded to `<pre>`
- Full test suite (`CodeBlock.vanilla.test.tsx`) covering copy behavior, fake-timer revert, accessibility (axe), ref forwarding, language variants, and theme support

### New: `useColorScheme` hook in `FaencyProvider`
- Reads `dark`/`light` class or `data-theme` attribute from `<html>`, falls back to `prefers-color-scheme`
- MutationObserver keeps the value reactive when the host app switches themes
- Storybook `preview.jsx` now toggles those classes on `<html>` so the hook resolves correctly in stories

### Bug fixes
- **Build**: marked all `@vanilla-extract/*` packages as external in Vite config so they aren't bundled into the library output
- **List**: fixed nested list support — `interactive` now propagates via context to child `Ul`/`Ol`; `Li` separates nested list nodes from inline content so they render outside the button/span wrapper; CSS selectors tightened to direct-child (`>`) combinators to avoid leaking styles into nested levels
- **AriaTable**: replaced `NonNullable<unknown>` with an explicit `{ children?: React.ReactNode; css?: CSS }` intersection in `AriaTableProps`
- **Button**: added `boxShadow: 'none'` to the `secondary + ghost` compound variant to prevent an inherited shadow from bleeding through

Fix: https://github.com/traefik/hub-issues/issues/2323

## Dependency changes

| Dependency | Version | State |
| --- | --- | --- |
| `prism-react-renderer` | `2.4.1` | Added |

## Breaking changes

None.

## How to test?

- [ ] Storybook: verify CodeBlock renders in light and dark modes, copy button toggles state and reverts after 1.5 s
- [ ] Storybook: verify nested `Ul`/`Ol` lists render and respect `interactive` at all nesting levels
- [ ] `yarn test:ci` passes (CodeBlock vanilla tests cover the copy/timer behavior)
- [ ] `yarn build` produces no bundled `@vanilla-extract` code in the output

## Good PR checkboxes

- [x] Change has been tested
- [x] Added/Updated tests
- [x] Added/Updated stories
- [x] Labels are set
- [x] Project is linked

## Good Review checkboxes

<details>

```markdown
- [ ] I've tested the changes
```

</details>

### DIFF
--- a/components/CodeBlock/CodeBlock.tsx
+++ b/components/CodeBlock/CodeBlock.tsx
@@ -1,6 +1,6 @@
 import { CheckIcon, CopyIcon } from '@radix-ui/react-icons';
 import { Highlight, Language, themes } from 'prism-react-renderer';
-import React, { ComponentProps, useState } from 'react';
+import React, { ComponentProps, forwardRef, useState } from 'react';
 
 import { CSS, styled } from '../../stitches.config';
 import { AccessibleIcon } from '../AccessibleIcon';
@@ -11,7 +11,6 @@ import { useColorScheme } from '../FaencyProvider';
 const Pre = styled('pre', {
   m: 0,
   borderRadius: '$2',
-  maxHeight: 424,
   overflow: 'auto',
   border: '1px solid $colors$deepBlue3',
   position: 'relative',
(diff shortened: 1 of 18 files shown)


components/CodeBlock/CodeBlock.tsx
```diff
@@ -0,0 +1,111 @@
...
+  const systemScheme = useColorScheme();
+  const resolvedScheme = colorScheme ?? systemScheme;
+  const prismTheme = resolvedScheme === 'dark' ? themes.nightOwl : themes.nightOwlLight;
+
+  const [copied, setCopied] = useState(false);
+
+  const handleCopy = async () => {
+    await navigator.clipboard.writeText(code);
+    setCopied(true);
+    setTimeout(() => setCopied(false), 1500);
```
gndz07: WDYT of handling a promise rejection fallback here?

components/CodeBlock/CodeBlock.vanilla.tsx
```diff
@@ -0,0 +1,111 @@
...
+  lang: Language;
+  code: string;
+  copyable?: boolean;
+  copyText?: string;
+  copiedText?: string;
+  noBorder?: boolean;
+  wrapText?: boolean;
+}
+
+export const CodeBlockVanilla = forwardRef<HTMLPreElement, CodeBlockVanillaProps>(
```
gndz07, line 35: Is there any reason why this one uses `forwardRef` and not the stitches one?

components/CodeBlock/CodeBlock.vanilla.css.ts
```diff
@@ -0,0 +1,33 @@
...
+import { tokens } from '../../styles/tokens.css';
+
+export const wrapper = style({
+  position: 'relative',
+});
+
+export const pre = style({
+  margin: 0,
+  borderRadius: tokens.radii['2'],
+  maxHeight: 424,
```
gndz07: May I ask why this exact size? If it's required, maybe we could extract it as a variable?

components/CodeBlock/CodeBlock.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+export type CodeBlockProps = {
+  lang: Language;
+  code: string;
+  copyable?: boolean;
+  copyText?: string;
+  copiedText?: string;
+  noBorder?: boolean;
+  wrapText?: boolean;
+  colorScheme?: 'light' | 'dark';
```
gndz07, line 53: Why does this accept `colorScheme` override and not the vanilla counterpart?

components/CodeBlock/CodeBlock.vanilla.test.tsx
```diff
@@ -0,0 +1,215 @@
...
+    );
+    expect(container.querySelector('pre')).toBeInTheDocument();
+  });
+
+  it('should apply wrapText prop', () => {
+    const { container } = renderWithTheme(
+      <CodeBlockVanilla lang="typescript" code={CODE_SNIPPET} wrapText />,
+    );
+    expect(container.querySelector('pre')).toBeInTheDocument();
+  });
```
gndz07: These 2 tests are identical in the sense that they just checked whether or not the pre component is present, it's not giving plus value in the sense of checking whether or not `noBorder` or `wrapText` is applied correctly. I think we could just remove it or update it accordingly.

components/CodeBlock/CodeBlock.vanilla.test.tsx
```diff
@@ -0,0 +1,215 @@
...
+
+  it('should work with dark theme', () => {
+    const { container } = render(
+      <VanillaExtractThemeProvider defaultTheme="dark">
+        <CodeBlockVanilla lang="typescript" code={CODE_SNIPPET} />
+      </VanillaExtractThemeProvider>,
+    );
+    expect(container.querySelector('pre')).toBeInTheDocument();
+  });
```
gndz07: Idem with previous comment, no particular check is done to validate against the custom props.

components/FaencyProvider.tsx
```diff
@@ -1,10 +1,63 @@
...
+  if (el.classList.contains('dark')) return 'dark';
+  if (el.classList.contains('light')) return 'light';
+  const attr = el.dataset.theme;
+  if (attr === 'dark' || attr === 'light') return attr;
+  return null;
+}
+
+const ColorSchemeContext = createContext<ColorScheme>('light');
+
+function ColorSchemeProvider({ children }: { children: React.ReactNode }) {
```
gndz07: Nit; I think it'd be nice to have these in a separate file.

components/CodeBlock/CodeBlock.tsx
mdeliatf: I'll include that, thanks for pointing it out!

components/CodeBlock/CodeBlock.vanilla.tsx
mdeliatf: I think it’s a mistake not to have it, I'll add it. Thanks for catching that!

components/CodeBlock/CodeBlock.vanilla.css.ts
mdeliatf: There’s no particular reason, it just felt right, but I agree it should be included as a prop!

components/CodeBlock/CodeBlock.tsx
mdeliatf: I think it’s a mistake not to have it, I'll add it. Thanks for catching that!

components/FaencyProvider.tsx
```diff
@@ -1,56 +1,8 @@
...
-    };
-  }, []);
-
-  return <ColorSchemeContext.Provider value={scheme}>{children}</ColorSchemeContext.Provider>;
-}
-
-export function useColorScheme(): ColorScheme {
-  return useContext(ColorSchemeContext);
-}
+export { useColorScheme } from './colorSchemeContext';
```
gndz07: IMHO it'd be better to just export it directly from index.ts

components/colorSchemeContext.tsx
```diff
@@ -0,0 +1,52 @@
+import React, { createContext, useContext, useEffect, useState } from 'react';
+
+export type ColorScheme = 'light' | 'dark';
+
+export function readDocumentTheme(): ColorScheme | null {
```
gndz07:
```suggestion
function readDocumentTheme(): ColorScheme | null {
```

components/colorSchemeContext.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const el = document.documentElement;
+  if (el.classList.contains('dark')) return 'dark';
+  if (el.classList.contains('light')) return 'light';
+  const attr = el.dataset.theme;
+  if (attr === 'dark' || attr === 'light') return attr;
+  return null;
+}
+
+export const ColorSchemeContext = createContext<ColorScheme>('light');
```
gndz07:
```suggestion
const ColorSchemeContext = createContext<ColorScheme>('light');
```

gndz07 (approved):
LGTM

PR #5134, opened 2026-04-20
feat: various CodeBlock improvements, fixes, and additional languages

mdeliatf:
## Description

This PR improves the `CodeBlock` component with extended language support, more flexible copy button positioning, and several bug fixes.

**Extended language support**: Introduces `prism-extend.ts` to register additional Prism.js grammars (`bash`, `csharp`, `docker`, `hcl`, `http`, `ini`, `java`, `php`, `powershell`, `protobuf`, `ruby`, `scala`, `shell-session`, `toml`) not included by default in `prism-react-renderer`. The `CodeBlockLanguage` type is updated to reflect all supported languages. Unknown languages fall back to `text` gracefully.

**Copy button alignment overhaul**: The `copyButtonAlign` prop is extended from a two-value `'top' | 'bottom'` to a richer `CopyButtonAlign` type (`'top' | 'top right' | 'bottom' | 'bottom right' | 'bottom left'`). The button position now adjusts dynamically via `ResizeObserver` to account for scrollbar width/height, preventing overlap.

**New props**: `onCopy` callback and `copyButtonBgColor` are added to both the Stitches and vanilla-extract implementations.

**Bug fix**: Code block background is set to `inherit` instead of `transparent` to correctly pick up the theme background.

## Preview

No visual changes.

## Dependency changes

| Dependency | Version | State |
| ---------- | ------- | ----- |
| `prismjs` | 1.30.0 | Added |
| `@types/prismjs` | 1.26.6 | Added |

## Breaking changes

None.

## How to test?

- [ ] Verify all new languages render with syntax highlighting in Storybook
- [ ] Verify fallback to `text` when an unsupported language is passed
- [ ] Verify copy button positions (`top right`, `bottom right`, `bottom left`) with and without scrollbars
- [ ] Verify `onCopy` callback fires on click
- [ ] Verify `copyButtonBgColor` applies correctly
- [ ] Run `yarn test:ci` — all tests pass

## Good PR checkboxes

- [ ] Change has been tested
- [ ] Added/Updated tests
- [ ] Added/Updated stories
- [ ] PR follows [conventions](https://github.com/traefik/faency#how-to-contribute)
- [ ] Labels are set
- [ ] Project is linked

## Good Review checkboxes

<details>

```markdown
- [ ] I've tested the changes
```

</details>


components/CodeBlock/CodeBlock.tsx
```diff
@@ -77,53 +93,112 @@ export const CodeBlock = forwardRef<HTMLPreElement, CodeBlockProps>(
...
+      </Button>
+    ) : null;
+
     return (
-      <Highlight theme={prismTheme} code={code} language={lang}>
-        {({ className, style, tokens, getLineProps, getTokenProps }) => (
+      <Highlight
+        theme={prismTheme}
+        code={code}
+        language={(lang != null && Prism.languages[lang] != null ? lang : 'text') as Language}
```
gndz07: Nit; this component defaults lang to 'text' already so a nul lang is not possible, isn't it?
```suggestion
        language={(Prism.languages[lang] != null ? lang : 'text') as Language}
```

components/CodeBlock/CodeBlock.vanilla.tsx
```diff
@@ -57,71 +72,128 @@ export const CodeBlockVanilla = forwardRef<HTMLPreElement, CodeBlockVanillaProps
...
+      : copyButtonWrapperTop;
+
+    const copyButton = copyable ? (
+      <ButtonVanilla
+        variant="secondary"
+        ghost
+        size="small"
+        type="button"
+        onClick={handleCopy}
+        css={{ color: '$hiContrast', gap: '$1', backgroundColor: copyButtonBgColor }}
```
gndz07: I have mixed feelings about passing `copyButtonBgColor` as a prop. WDYT of having it as a token, so that if needed user can override the color by re-declaring the color variable of the token?

components/CodeBlock/CodeBlock.vanilla.tsx
```diff
@@ -1,31 +1,44 @@
...
+  useLayoutEffect,
+  useRef,
+  useState,
+} from 'react';
 
 import { CSSProps, processCSSProp } from '../../styles/cssProps';
 import { useVanillaExtractTheme } from '../../styles/themeContext';
 import { AccessibleIcon } from '../AccessibleIcon';
 import { ButtonVanilla } from '../Button';
+import { CopyButtonAlign } from './CodeBlock';
```
gndz07: I wouldn't import it from the stiches component file, it's vulnerable to future mistake when we're fully migrating. Maybe we can move CopyButtonAlign to index.

components/CodeBlock/CodeBlock.tsx
```diff
@@ -26,28 +35,31 @@ const Pre = styled('pre', {
...
   zIndex: 1,
-
   variants: {
-    align: {
-      top: { top: '$2' },
-      bottom: { bottom: '$2' },
+    position: {
+      top: { top: '$2', right: '$3' },
+      bottomRight: { bottom: '$2', right: '$3' },
+      bottomLeft: { bottom: '$2', left: '$3' },
```
gndz07, line 44: Why bottom has right and left but not top?

gndz07 (changes requested):

components/CodeBlock/CodeBlock.tsx
mdeliatf: There’s no specific reason, it’s simply that we haven’t encountered a use case for it yet, and it would overlap the code. That said, we could handle it the same way we do at the bottom by adding a top margin!

gndz07: For the sake of usage clarity IMHO it's better to keep it mirrored (either to have left/right sides or not, but to keep it the same for top and bottom). Otherwise, it should be well documented that top and bottom have different expected values.

gndz07 (approved):
LGTM

